PR #209, opened 2026-05-10
Retire deprecated orchestrated_release_prep + scaffolding (v7.0.0)

silversurfer562:
## Summary

Retires two long-deprecated production modules per `docs/specs/deprecated-module-retirement/`. **Breaking change — slated for v7.0.0**, which couples with the upcoming 100%-coverage milestone release.

- **`attune.workflows.orchestrated_release_prep`** (637 lines). Deprecated v5.2.0, scheduled for removal in v6.0; we are at v6.6.0 — six minor versions overdue. Migration: `ReleasePrepTeamWorkflow` from `attune.agents.release` (drop-in replacement).
- **`attune.scaffolding`** (entire CLI package, 9 files / 2,254 lines). Deprecated 2026-02-21 (PR #60); `__main__.py` has emitted a runtime deprecation notice on every invocation. Migration: `attune workflow run <workflow-name>`.

Cross-repo grep confirmed zero callers across the eight sibling `attune-*` repos. Full per-commit detail in `docs/specs/deprecated-module-retirement/decisions.md`.

Side cleanups picked up while removing:
- `examples/orchestration/basic_usage.py`: rewrote release-prep examples against the new API; removed broken Example 3 (`test_coverage_boost` was already deleted in a prior release, leaving a dead import).
- Trimmed deprecated test classes from `tests/unit/test_coverage_batch6.py` and `tests/unit/workflows/test_workflow_consolidation.py`.
- Removed the `python -m attune.scaffolding` row from `docs/reference/cli-reference.md`.
- Bug log: introduced **Class 5** (\"deprecated production code outliving its tests\") in `docs/COVERAGE_BUG_LOG.md`.

## Test plan

- [x] Full unit suite under `-n auto`: **14,103 passed, 0 failed** (baseline 14,110; -7 from deleted deprecated tests)
- [x] Sanity: `import attune.workflows.orchestrated_release_prep` → `ModuleNotFoundError`
- [x] Sanity: `import attune.scaffolding` → `ModuleNotFoundError`
- [x] Sanity: `from attune.workflows import ReleasePrepTeamWorkflow` succeeds and is identity-equal to canonical `attune.agents.release.ReleasePrepTeamWorkflow`
- [x] Cross-repo grep across 8 sibling `attune-*` repos: zero hits
- [x] Pre-commit hooks (black, ruff, bandit, secrets) green on all 4 commits
- [ ] CI parity check on this PR (pending CI run)
[...]

PR #212, opened 2026-05-10
ci: stabilize CI (canonical coverage + pytest-timeout + integration filter + test pollution fix)

silversurfer562:
## Summary

Three coordinated changes that together unblock CI on \`main\`. Originally drafted as PR #212 + PR #214, then combined here because each fix in isolation can't get a green CI run on top of the broken main.

The chain of root causes (each one only visible after the previous is fixed):

1. **Workers crashing silently** → \`pytest-timeout\` would name them
2. **Integration tests crash workers in CI** → exclude them with marker filter
3. **\`token_estimator\` tests pollute \`sys.modules\`** → use \`monkeypatch.delitem\` so cleanup is automatic

Three commits, one per fix, granular history preserved.

---

## Commit 1 — \`pytest-timeout\` for diagnostic value (3f3b9a66)

Adds \`pytest-timeout>=2.3.0,<3.0\` to \`[dev]\` and passes \`--timeout=60 --timeout-method=thread\` to the CI pytest invocation. Without this, hanging tests killed xdist workers silently — manifesting as partial runs + bogus low coverage. With it, a hung test fails with a stack trace pointing at the exact line.

- 60s ceiling — slowest known unit test is ~3s; 20x headroom for genuinely slow tests
- \`thread\` method (not default \`signal\`) for cross-platform support; \`SIGALRM\` is unreliable on Windows

## Commit 2 — Skip integration-marked tests in CI gate (cb95af9c)

Commit 1 immediately revealed the source of the silent failures: 4 tests in \`tests/unit/orchestration/test_execution_strategies.py\`, class \`TestIntegrationWithRealAgents\`, crashing their workers in CI.

The class is already correctly tagged \`@pytest.mark.integration\` and \`@pytest.mark.slow\`; the docstring explicitly says *\"Run with: pytest -m integration / Skip with: pytest -m 'not integration'\"*. CI just wasn't filtering them — \`-m \"not network\"\` only excludes a different marker.

Fix: extend the marker filter to \`not network and not integration\`. Verified locally:
\`\`\`
pytest tests/unit/orchestration/test_execution_strategies.py \\
  -m \"not integration\" --collect-only
→ 89/93 tests collected (4 deselected) — exactly the 4 crashing classes excluded
\`\`\`

Two other \`tests/unit/\` files also have \`@pytest.mark.integration\` classes and benefit from the same exclusion: \`test_security_remediation.py:308\` and \`test_routing_and_cost.py:829\`.

## Commit 3 — Fix \`sys.modules\` pollution in \`token_estimator\` tests (94c11853)

Commit 2 made tests reach 100%, which then exposed a third bug: 2 tests in \`tests/unit/test_token_utils.py\` (\`test_api_counting_success\`, \`test_api_message_counting\`) failing only under xdist + only when run alongside \`tests/unit/models/test_token_estimator.py\`.

Root cause: 3 tests in \`test_token_estimator.py\` do bare \`sys.modules.pop(\"attune.utils.tokens\", None)\` to test ImportError fallbacks but never restore. Subsequent tests on the same xdist worker that imported \`count_tokens\` at file-collection-time end up with a function whose \`__globals__\` points to the now-popped module — bypassing their \`@patch(\"attune.utils.tokens._get_client\")\` decorator entirely. The real production env-var check fires; the assertion fails.

Fix: replace the 3 \`sys.modules.pop\` calls with \`monkeypatch.delitem(sys.modules, \"attune.utils.tokens\", raising=False)\`. Both forms remove the cached module so the test's \`fake_import\` side_effect can fire on the next import attempt; \`monkeypatch\` additionally restores at test teardown, killing the cross-test leak.

3 lines changed, one per test.

---

## Test plan

- [x] Locally verified (full unit suite under \`-n auto\`):
  - Before: 5 failed, 14,122 passed, ~62s
  - After all 3 commits: 3 failed, 14,124 passed, ~50s (+2 tests recovered, suite is faster too)
  - The 3 remaining failures are unrelated env/path flakes
- [x] Sanity check on commit 2: \`pytest -m \"not integration\" --collect-only\` deselects exactly the 4 crashing tests
- [x] Sanity check on commit 3: the 3 modified tests still pass in isolation and as a group
- [ ] CI on this combined PR confirms cross-platform pass

## Bug-log classification

- Commit 1: tooling, not a bug per se
- Commit 2: misfiled tests (integration tests in \`tests/unit/\`), exposed by Phase A's expanded \`[dev]\` extras
- Commit 3: **Class 3 sub-pattern** — \"tests that mutate global module state without restoring it.\" Distinct from canonical Class 3 (\"tests mocking around the bug\") because here production code is correct; the bug is in test setup leaking into other tests' assertion paths.

## Notes

- Supersedes #214 (token_estimator pollution fix as a standalone PR) — closing that.
- Closes the silent-CI-failure regression from PR #207 (Phase A).
- Once this lands, branch-protection-friendly CI returns. PRs #209 and #213 can rebase and merge cleanly afterward.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -51,7 +51,7 @@ jobs:
       # they exercise but are not part of the standard CI gate. Run them
       # separately via `pytest -m integration` against a configured env.
       run: |
-        pytest --cov=src/attune --cov-branch --cov-report=xml --cov-report=term-missing --cov-fail-under=85 --timeout=60 --timeout-method=thread -m "not network and not integration"
+        pytest --cov=src/attune --cov-branch --cov-report=xml --cov-fail-under=85 --timeout=60 --timeout-method=thread -m "not network and not integration"
       env:
         ANTHROPIC_API_KEY: ${{ secrets.ANTHROPIC_API_KEY }}
 
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -77,7 +77,7 @@ jobs:
           MEM_MONITOR_PID=$!
           trap 'kill $MEM_MONITOR_PID 2>/dev/null || true' EXIT
         fi
-        pytest -n 2 --cov=src/attune --cov-branch --cov-report=xml --cov-fail-under=85 --timeout=60 --timeout-method=thread -m "not network and not integration"
+        pytest -n 2 --cov=src/attune --cov-report=xml --cov-fail-under=85 --timeout=60 --timeout-method=thread -m "not network and not integration"
         if [ "${{ runner.os }}" = "Linux" ]; then
           echo "[mem-final $(date +%H:%M:%S)] $(free -m | awk '/Mem:/ {print "used="$3"MB avail="$7"MB"}')"
         fi
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -39,8 +39,13 @@ jobs:
         pip install -e .[dev]
 
     - name: Run tests with coverage
+      # --timeout=60 + --timeout-method=thread: catches a hanging test by
+      # name with a stack trace instead of letting it kill the xdist worker
+      # silently (which manifests as partial test runs + bogus low coverage).
+      # `thread` method is required for cross-platform support (default
+      # `signal` uses SIGALRM, which is unreliable on Windows).
       run: |
-        pytest --cov=src/attune --cov-branch --cov-report=xml --cov-report=term-missing --cov-fail-under=85 -m "not network"
+        pytest --cov=src/attune --cov-branch --cov-report=xml --cov-report=term-missing --cov-fail-under=85 --timeout=60 --timeout-method=thread -m "not network"
       env:
         ANTHROPIC_API_KEY: ${{ secrets.ANTHROPIC_API_KEY }}
 
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -44,8 +44,14 @@ jobs:
       # silently (which manifests as partial test runs + bogus low coverage).
       # `thread` method is required for cross-platform support (default
       # `signal` uses SIGALRM, which is unreliable on Windows).
+      #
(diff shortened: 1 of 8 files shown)


silversurfer562:
## Probe B finding (run 25643234935, ubuntu-3.11)

Memory instrumentation **confirms the OOM hypothesis**:

| Time | Used | Available | Note |
|---|---|---|---|
| 00:00:12 | 15,623 MB | 366 MB | tests running |
| 00:00:42 | 15,698 MB | 291 MB | continuing |
| 00:01:12 | 15,738 MB | 251 MB | brink of 16 GB ceiling |
| **00:01:42** | **1,393 MB** | **14,595 MB** | **worker killed → 14 GB reclaimed** |
| 00:02:12 | 2,260 MB | 13,729 MB | suite resumes on remaining workers |

The `[100%] PASSED → runner shutdown` pattern was the Linux OOM killer harvesting an xdist worker at the 16 GB cap. The test marked `FAILED` at 00:01:42 (`test_clear_session_without_memory`) is the casualty — not the cause; it was simply on the worker that got killed.

**Root cause (most likely):** pytest-cov keeps per-worker coverage data in-memory across the suite. At ~14k tests × `-n auto` (4-6 workers on Linux runners) × branch coverage, this exceeds 16 GB.

**Mitigations to explore (Probe C territory):**
- Cap xdist workers explicitly (`-n 2` or `-n 4`) instead of `-n auto`
- Drop `--cov-branch` (cheaper) and/or split coverage across multiple smaller pytest runs
- Use `coverage` with a context-aware data file rather than in-memory accumulation
- Run tests in two passes: tests-only (no cov) + coverage-only on a subset

This PR's 3 commits (timeout, integration filter, pollution fix) are independent quality wins and don't depend on the OOM fix. Probe B's mem-instrumentation is also valuable to keep merged so future regressions are visible.

Suggest: merge this once Windows/macOS confirm the shell/lock fixes, file OOM mitigation as separate Probe C work.

PR #441, opened 2026-05-19
feat(ops): home account-spend tiles (anthropic-cost-integration Phase 2)

silversurfer562:
## Summary

Wires Phase 1's `anthropic_cost.fetch_summary` into the home page route. Adds three account-spend KPI tiles ("Today (account)", "7d (account)", "MTD (account)") gated on cost_summary availability, with a "Connect billing" CTA when no admin key is configured and a friendly "Key rejected" notice on auth-failed.

Implements spec [`docs/specs/anthropic-cost-integration/tasks.md`](docs/specs/anthropic-cost-integration/tasks.md) Phase 2.

## What's in the diff

- **`src/attune/ops/routes/dashboard.py`**: home() now calls fetch_summary inside a defensive try/except, passes cost_summary and cost_error to the template; ?refresh=1 forces a cache bypass
- **`src/attune/ops/templates/home.html`**: new `<section class="kpi-grid kpi-cluster-account">` block after the existing telemetry cluster, with three render paths (valid summary → tiles, no_key → CTA, auth_failed → notice, other errors → silent fall-through)
- **`src/attune/ops/static/css/main.css`**: `.kpi-cluster-account` styles, including the thin left-edge accent border and color-coded variants for `.kpi-cta` and `.kpi-notice`
- **`tests/unit/ops/test_anthropic_cost_home_integration.py`**: six new tests covering all four spec-2.4 scenarios plus the default-refresh-false and exception-swallow paths

## Trade-offs / non-obvious choices

- **CTA copy**: shipped option 1 from the open-question list (`"Set up cost reporting"` linking to docs/reference/) rather than option 2 (`"Coming in Phase 4 …"`). Cleaner, doesn't reference a future state. Easily revisited when Phase 4 lands.
- **Visual distinction**: left-edge accent border (3px solid var(--accent)) rather than the spec's "right-edge" suggestion. Reads more naturally with the existing tile flex direction; functionally equivalent for the "tell at a glance which data source" goal.
- **Defensive try/except**: catches everything (`except Exception` with `# noqa: BLE001`) because a billing-fetch surprise must not block the dashboard render. Surfaces in DEBUG logs only — verified by the silent-degradation test which mocks fetch_summary to raise RuntimeError.
- **Patch target in tests**: `attune.ops.anthropic_cost.fetch_summary` (source module) rather than `attune.ops.routes.dashboard.anthropic_cost.fetch_summary` (consumer site). The latter triggers `mock.patch`'s `_importer` to attempt module-walk on a function attribute. Source-module patching works because `dashboard.anthropic_cost` IS the same module object as the source; replacing the attribute on the source flows through.

## Caught in flight

The formatter stripped the `from attune.ops import anthropic_cost` import after my initial import-only edit (before the usage line was added), per the existing CLAUDE.md lesson about formatters stripping unused-at-the-moment imports. Re-added with the usage line in place; runtime confirmed `dashboard.anthropic_cost` resolves to the module. Worth a session lesson: even brief edit-formatter cycles trigger this, not just multi-minute pauses.

## Test plan

- [x] 6 new tests in `test_anthropic_cost_home_integration.py` pass
- [x] 50 tests in `test_home.py` + `test_anthropic_cost.py` + new file pass (no regressions)
- [x] Ruff clean on Python files
- [x] Pre-commit hooks pass (black, ruff, bandit, detect-secrets, markdown checks, help-template freshness)
- [ ] CI green on all platforms
- [ ] Manual verification: dashboard launches with no admin key → CTA tile renders; with valid CostSummary mock → three tiles populated

## Related

- [#432](https://github.com/Smart-AI-Memory/attune-ai/pull/432) — Phase 1 backend client this PR consumes
- [#431](https://github.com/Smart-AI-Memory/attune-ai/pull/431) — original spec
- Next: Phase 3 (/telemetry per-day panel), Phase 4 (`attune ops setup-billing` CLI), Phase 5 (polish + docs)

### DIFF
--- a/src/attune/ops/routes/dashboard.py
+++ b/src/attune/ops/routes/dashboard.py
@@ -8,7 +8,7 @@ from pathlib import Path
 from fastapi import APIRouter, Request
 from fastapi.responses import HTMLResponse
 
-from attune.ops import data
+from attune.ops import anthropic_cost, data
 
 logger = logging.getLogger(__name__)
 
@@ -60,6 +60,20 @@ async def home(request: Request) -> HTMLResponse:
     runner = getattr(request.app.state, "runner", None)
     recent_runs = [r.to_dict() for r in runner.recent(limit=5)] if runner else []
     attune_ai = next((v for v in versions if v.package == "attune-ai"), None)
+    # Anthropic account-spend tiles (Phase 2 of anthropic-cost-integration).
+    # Defensive try/except — a fetch crash must not block the dashboard
+    # render, so we degrade silently (no tile cluster) on unexpected
+    # errors. Categorized failures (no_key, auth_failed, rate_limited,
+    # network) come back through cost_error and the template renders
+    # the right CTA / notice / fallback per-kind.
+    refresh = request.query_params.get("refresh") == "1"
+    try:
+        cost_summary, cost_error = anthropic_cost.fetch_summary(refresh=refresh)
+    except Exception:  # noqa: BLE001
+        # INTENTIONAL: defensive degradation; surface in DEBUG logs but
+        # never block the home page render on a billing-fetch surprise.
+        logger.debug("anthropic_cost.fetch_summary raised", exc_info=True)
+        cost_summary, cost_error = None, None
     return _render(
         request,
         "home.html",
@@ -71,6 +85,8 @@ async def home(request: Request) -> HTMLResponse:
         sparkline=sparkline,
         recent_runs=recent_runs,
         attune_ai=attune_ai,
+        cost_summary=cost_summary,
+        cost_error=cost_error,
     )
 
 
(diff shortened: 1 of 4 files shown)


silversurfer562:
Rebased onto current main (which now includes #439 telemetry fix and #440 docs). The earlier diff against `origin/main` showed -300 lines because the branch was behind main — what looked like a bundled removal of `_track_sdk_run_telemetry` and friends was actually just absence of #439's additions on this branch.

After rebase, the diff is exactly 4 files (+255 / -1):
- `src/attune/ops/routes/dashboard.py` — home route wires `anthropic_cost.fetch_summary`
- `src/attune/ops/templates/home.html` — 3 KPI tiles + no_key CTA + auth_failed notice
- `src/attune/ops/static/css/main.css` — `.kpi-cluster-account` styling
- `tests/unit/ops/test_anthropic_cost_home_integration.py` — 6 integration tests

40/40 tests pass post-rebase. PR is now cleanly aligned with what the title advertises.

PR #745, opened 2026-06-10
feat(voice): workflow-result-formatting T3 — WorkflowReport wiring + safety net

silversurfer562:
## Summary

Workflow-result-formatting **T3** — the voice layer now consumes the T1 data model (#649) + T2 renderer (#741):

- **`WorkflowReport` detection** (design D2): `_extract_from_workflow_result` checks `final_output` for the `_type: "WorkflowReport"` discriminator before the `str()` fallback, reconstructs via `from_dict`, and renders through `report_renderer.render_safe(disclosure="summary", show_cost=resolve_show_cost())`. Score comes from the report; the formatter's own cost line is suppressed for rendered reports (the renderer owns cost — no double display). A rendered report is authoritative: never overridden by the sparse-text fallback. Malformed payloads log + degrade to plain-dict handling, never crash.
- **Safety net** (proposal §6): the bare `str(final_output)` branch is now a visible `⚠ Renderer not yet migrated for <Type>` banner + repr-free pretty-printer (enums → `.value`, containers → counts, nested dataclasses → type name). Plain-string `final_output` passes through untouched — that's a normal text report, not a repr leak.
- **`resolve_show_cost()` + `AttuneConfig.show_cost_metrics`** (design D3): tri-state — explicit config wins; `None` = auto (on for API-key users, off for subscription users; empty `ANTHROPIC_API_KEY` counts as absent, matching CI). `ATTUNE_SHOW_COST_METRICS` parsed in `from_env`. Config-load failure degrades to auto, never breaks output.

## Test plan

- 26 new tests in `tests/unit/voice/test_formatter_report_wiring.py` (report branch, safety net, show-cost resolution, config field/env parsing)
- All 91 pre-existing voice tests + 266 config tests + MCP/ops/CLI formatter consumers pass unchanged
- `attune.voice.formatter` at 94% branch coverage; every remaining miss is a pre-existing line outside this diff

Spec: `docs/specs/workflow-result-formatting` (tasks.md updated; T4 CLI is next)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- **Voice wiring for structured workflow reports (workflow-result-formatting
+  T3).** `_extract_from_workflow_result` now detects a serialized
+  `WorkflowReport` (`_type` discriminator) in `final_output` and renders it
+  through `attune.voice.report_renderer.render_safe` (summary disclosure);
+  the report's score travels through, and the formatter's own cost line is
+  suppressed for rendered reports (the renderer owns cost). New
+  `resolve_show_cost()` + `AttuneConfig.show_cost_metrics` (tri-state:
+  explicit override, else auto — on for API-key users, off for
+  subscription users; design D3). The bare `str(final_output)` branch is
+  now a safety net: a visible "Renderer not yet migrated for X" banner
+  plus a repr-free field pretty-printer (proposal §6). Plain-string
+  `final_output` still passes through untouched.
+
 ## [8.2.0] — 2026-06-10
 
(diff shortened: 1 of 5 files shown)


silversurfer562:
Closing as superseded: this was a parallel session's T3, and the same task merged first as #746 (`8a916a6f`) — hence the conflicts on every touched file. Rather than rebase this down to a duplicate, its two genuine improvements over the merged version were ported to the open T4 PR #749 (commit `76aec07c`, with tests):

1. **Malformed-report hardening** — `WorkflowReport` reconstruction failure now degrades to legacy dict handling instead of a `str(WorkflowResult)` repr leak.
2. **Voice cost-line suppression for rendered reports** — the renderer's `show_cost` gate owns cost display (design D3); no wrapper duplicate, and no inapplicable cost line for subscription users.

Nothing else here was unique vs #746 (same config field, same env parsing, same safety-net shape). Thanks, other me.

PR #861, opened 2026-06-13
fix(platform): remove dead setup_asyncio_policy that set the wrong Windows loop policy

silversurfer562:
## Problem

`platform_utils.setup_asyncio_policy()` set `asyncio.WindowsSelectorEventLoopPolicy()` on Windows, with a docstring claiming it helps "particularly with subprocesses." That's **backwards** — on Windows, `SelectorEventLoop` *cannot* spawn subprocesses (`create_subprocess_exec` raises `NotImplementedError`); only `ProactorEventLoop` can, and Proactor is already the Windows default. The function installed a broken, non-default policy **process-wide**.

It was also **dead code** with **harmful side effects**:
- `setup_asyncio_policy()` was called only by `safe_run_async()`, which has **zero production callers** (neither is in `__all__`/`__init__`; no sibling/external use).
- `scripts/check_platform_compat.py` emitted an `info` advisory on *every* `asyncio.run()` recommending this broken function (~12 hits) — spreading the wrong pattern.
- Tests that called it directly **polluted the global event-loop policy**, which non-deterministically poisoned the help-regen subprocess tests on Windows CI (the root cause behind [#840](https://github.com/Smart-AI-Memory/attune-ai/pull/840)'s skips).

## Change

- Remove `setup_asyncio_policy()` and `safe_run_async()` from `platform_utils.py`.
- Drop the `asyncio_policy` rule from `check_platform_compat.py` (+ its `ASYNCIO_RUN` pattern and docstring bullet).
- Delete the tests that exercised them.
- Update the #840 help-regen skip comment to its now-accurate reason (the fixture's `.bat` wrapper isn't directly launchable via `create_subprocess_exec` on Windows — the policy-pollution explanation no longer applies).

**Net −120 lines. No production behavior change** — the removed code was unreachable.

## Note

This removes two undocumented, non-`__all__` module-level functions. Any external code importing `attune.platform_utils.{setup_asyncio_policy,safe_run_async}` would break — but both were unused internally and one installed a broken policy, so keeping them is worse. Worth a "Removed" CHANGELOG line in the next release.

## Verification

- 129 passed / 1 skipped across the 4 affected test files (solo + xdist)
- `platform-compat` CI command exits 0 (0 errors); no more asyncio advisories
- ruff + black clean on all 6 files

### DIFF
--- a/scripts/check_platform_compat.py
+++ b/scripts/check_platform_compat.py
@@ -4,7 +4,6 @@
 Scans the codebase for common cross-platform issues:
 - Hardcoded Unix paths (/var/log, /tmp, etc.)
 - open() calls without encoding specified
-- asyncio.run() without setup_asyncio_policy()
 - os.path operations that should use pathlib
 
 Usage:
@@ -79,8 +78,6 @@ OPEN_WITHOUT_ENCODING = re.compile(
     re.MULTILINE,
 )
 
-ASYNCIO_RUN = re.compile(r"\basyncio\.run\s*\(")
-
 OS_PATH_OPERATIONS = [
     (r"\bos\.path\.join\s*\(", "Consider using pathlib.Path instead of os.path.join"),
     (r"\bos\.path\.exists\s*\(", "Consider using Path.exists() instead"),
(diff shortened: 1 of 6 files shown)


silversurfer562:
Closing in favor of #847 (parallel session) — same setup_asyncio_policy fix, chosen by the owner. #847 keeps the API surface and adds a conftest policy-restore guard.

PR #882, opened 2026-06-14
test(ci): Phase 4 out-of-class verification (src edit, must NOT auto-merge)

silversurfer562:
Phase 4 verification of the auto-merge-safe class — **out-of-class** case (comment-only `src/` edit; coverage stays green). Expected: NO `auto-merge-safe` label, and the `merge` job skips on its path-class re-check even when coverage is green. **This PR must NOT auto-merge** — it will be closed after verification.

### DIFF
--- a/src/attune/__init__.py
+++ b/src/attune/__init__.py
@@ -340,3 +340,6 @@ __all__ = [
     "get_redis_memory",
     "load_config",
 ]
+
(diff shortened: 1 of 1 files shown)


silversurfer562:
Throwaway negative-case test for the auto-merge-safe Phase-4 receipt — correctly stayed unlabeled/unmerged (out-of-class: touches src/). Closing now that the receipt is captured (#881 auto-merged in-class).

PR #885, opened 2026-06-14
fix(ci): harden auto-merge-safe sha→PR lookup (prefer event payload) [D6]

silversurfer562:
## What

Harden the `auto-merge-safe` merge job's SHA→PR resolution. Owning
spec: `docs/specs/auto-merge-safe-class/` (built #878, trigger fixed
#883). This is D6.

## Bug (PR #884, 2026-06-14)

With the #883 `workflow_run` trigger live, #884's `Tests` completion
DID invoke the merge job, but it logged `No open PR against main for
<sha>` and bailed — `gh api commits/$SHA/pulls` returned EMPTY while
#884 was plainly open against `main` with that exact head. The class
had **never merged a real PR end-to-end**.

## Root cause — verified

Eventual-consistency lag in the `commits/{sha}/pulls` association
index (NOT a PAT-visibility quirk). Natural experiment: SHAs
`2d9493ae` (#881) and `f904844d` (#873) each logged "No open PR"
inside the merge run, yet the identical query returns them as `open`
hours later while the PRs were open the whole time. A repo-scoped
fine-grained PAT has no per-PR visibility restriction for own-repo
PRs, so (b) is ruled out on mechanism.

## Fix

- **Primary** SHA→PR source: `github.event.workflow_run.pull_requests[]`
  — in the event payload, no indexing lag, populated for same-repo
  PRs (the class already requires head repo == base repo).
- **Fallback**: `commits/{sha}/pulls` retried with backoff (6×30s,
  10-min job timeout) for the rare empty-payload case.
- **Open + base==main re-checks moved into the per-PR loop**, so
  correctness is independent of which source resolved the PR. All D1
  gates (author, draft, fork, label, path-class re-check, coverage on
  head) preserved.
- Diagnostic logging of the raw payload + resolution source, so the
  Phase-4 run records empirically whether `workflow_run.pull_requests[]`
  populates on this repo.

Out of scope (noted in D5): `workflow_run` fires only when the WHOLE
`Tests` matrix completes, so a genuinely hung lane still delays merge.

## Decisions

`decisions.md` gains **D6** (this change) and **D5** (the
already-shipped #883 `check_run`→`workflow_run` trigger fix, recorded
for a gap-free history). @silversurfer562 / #883 owner: flagging the
D5 addition in case you intended to record it separately.

## Test plan

- [x] 33 path-guard unit tests pass (`auto_merge_guard.py` unchanged)
- [x] YAML validates; pre-commit clean
- [ ] **Phase 4 receipt (after this lands on main):** a tests/docs-only
      PR auto-merges; a `src/`-touching PR does NOT. (The merge job
      runs the workflow def from the default branch, so the fix must be
      on `main` before re-testing.)

### DIFF
--- a/.github/workflows/auto-merge-safe.yml
+++ b/.github/workflows/auto-merge-safe.yml
@@ -93,7 +93,7 @@ jobs:
   merge:
     if: github.event_name == 'workflow_run'
     runs-on: ubuntu-latest
-    timeout-minutes: 5
+    timeout-minutes: 10
     permissions:
       contents: read
     steps:
@@ -104,6 +104,10 @@ jobs:
         env:
           GH_TOKEN: ${{ secrets.ADMIN_MERGE_TOKEN }}
           SHA: ${{ github.event.workflow_run.head_sha }}
+          # PRIMARY sha->PR source: the workflow_run event payload's
+          # pull_requests[] (synchronous, no indexing lag, populated for
+          # same-repo PRs — and our class requires head repo == base repo).
+          WORKFLOW_RUN_PRS: ${{ toJSON(github.event.workflow_run.pull_requests) }}
           REPO: ${{ github.repository }}
           OWNER_LOGIN: silversurfer562
         run: |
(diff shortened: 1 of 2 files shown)


LuuOW (commented):
Technical audit: code patterns verified for system integrity.

PR #917, opened 2026-06-16
feat(deps): adopt claude-agent-sdk 0.2.x (lock 0.2.102)

silversurfer562:
## What

Deliberate adoption of `claude-agent-sdk` 0.2.x, per the
[claude-agent-sdk-0-2-migration spec](docs/specs/claude-agent-sdk-0-2-migration/).
Pin `>=0.2.101,<0.3.0` (lock at **0.2.102**); the prior deliberate
`<0.2.82` cap is lifted to a new `<0.3.0` guard against the next major.

## Why it's a clean bump (evidence)

**T0 — empirical scan against 0.2.102:**
- API-compatible: every symbol + `ClaudeAgentOptions` field attune
  uses still exists (no renames/removals reach attune).
- Full keyless unit suite: **17857 passed, 0 failed**, 72 skipped.

**T1 / T2 — static audit of the behavioral breaks → no code change:**
- *System-prompt default:* all 16 `ClaudeAgentOptions` sites splat
  `sdk_isolation_kwargs()` (`setting_sources=[]`), which already
  suppresses the Claude Code system prompt regardless of the SDK
  default. The existing drift-guard test enforces this.
- *MCP background-connection:* no workflow passes `mcp_servers`, so
  the default change doesn't affect runs.
- *TodoWrite→Task tools:* `TodoWrite` is unused in `src/`.

## Remaining validation

The keyless unit suite mocks the real SDK subprocess, so the live
loop still wants one real-SDK round-trip under 0.2.102. Recommend
running the budget-capped `integration-auth` workflow in CI's clean
env before merge (a local nested-Bash run would hit the SDK-nested
failure trap and give a false signal).


silversurfer562:
**Blocked: real-SDK break found by integration-auth.**

The mocked unit suite passed (17857), but the budget-capped `integration-auth` run on this branch [failed](https://github.com/Smart-AI-Memory/attune-ai/actions/runs/27591684930) — **systemically**, across all workflows (perf-audit, bug-predict, …):

```
Exception: Claude Code returned an error result: success
  claude_agent_sdk/_internal/query.py:852  receive_messages
```

The live SDK query path is broken under 0.2.102 — workflows return `source-failure` markers instead of analysis. This is the behavioral incompatibility the static T1/T2 audit could not see. Converting to draft; needs a real adapter/result-handling fix (or root-causing whether 0.2.x changed the result-message contract) before it can land.

PR #981, opened 2026-06-22
test(help): regression guard for manifest globs + template symbols

silversurfer562:
## What

Adds a regression guard for the `.help/` help system that catches two
drift classes the content-hash staleness system is structurally blind
to — the two the help-freshness sweep (#980) exposed.

`attune.help.staleness.compute_source_hash` / `check_staleness` hashes a
feature's **source** files. That flags edits to surviving files but
cannot see:

1. **Manifest globs referencing deleted files** — `features.yaml` can
   keep globbing a deleted path; the survivors' hash changes once, then
   looks "current" forever.
2. **Templates documenting deleted symbols** — the hash is over source,
   not templates, so a template can name a class that exists nowhere and
   staleness stays green.

## The guard

`tests/unit/help/test_help_manifest_integrity.py` — two hard assertions:

| Test | Scope | Check |
|------|-------|-------|
| `test_manifest_file_references_resolve` | all generated features | every `files` entry resolves — literal paths exist on disk, globs match ≥1 file (mirrors staleness's `**`→`**/*` + cache-dir exclusion) |
| `test_template_symbols_exist_in_source` | features with templates | every backticked `` `ClassName` `` / `` `func()` `` a feature's templates document appears in that feature's resolved source |

Symbol filtering: drops Python builtins + ALL-CAPS constants/env-vars,
plus a small `_IGNORED_SYMBOLS` (stdlib + verified cross-feature refs)
and a documented `_KNOWN_DEAD_SYMBOL_DRIFT` allow-list for pre-existing
out-of-scope drift.

## Drift the guard surfaced (and this PR fixes)

Three more dead manifest globs — same bug class as #980 — removed from
`features.yaml`. All hash-neutral (they matched zero files, so the
source hash and staleness are unchanged):

| Feature | Dead glob | Cause |
|---------|-----------|-------|
| `orchestration` | `src/attune/coordination/**` | module deleted in #279 |
| `code-quality` | `src/attune/workflows/code_review_*.py` | consolidated to `code_review.py` |
| `help-system` | `packages/attune-help/src/attune_help/**` | path not in this repo |

It also surfaced one dead symbol — `WorkflowExecutionError` in
`code-quality/error.md` (hallucinated, exists nowhere). Out of scope for
this fix-test-focused PR, so it's allow-listed with a comment and
tracked for a follow-up template rewrite.

## Coordination note — #980

PR #980's fix (`b50af7e0f`) is **not yet on main**. The template-symbol
check can only pass against the corrected fix-test templates, so this
branch **cherry-picks `b50af7e0f`** to stay self-consistent. Git handles
the duplicate cleanly when both merge — **sequence #980 first if
possible**, after which the cherry-pick here becomes a no-op on rebase.

## Verification

- `test_help_manifest_integrity.py`: **47 passed** (CI-faithful, `filterwarnings = error`)
- Full `tests/unit/help/`: **438 passed, 3 xfailed**
- ruff + pinned black: clean

### DIFF
--- a/.help/features.yaml
+++ b/.help/features.yaml
@@ -42,7 +42,6 @@ features:
     description: Review code for style issues, likely bugs, and structural problems
     files:
       - src/attune/workflows/code_review.py
-      - src/attune/workflows/code_review_*.py
     tags: [review, quality, bugs, lint]
 
   smart-test:
@@ -109,7 +108,6 @@ features:
     description: Progressive-depth help engine and template management
     files:
       - src/attune/help/**
-      - packages/attune-help/src/attune_help/**
     tags: [help, templates, docs]
     doc_paths:
       - docs/how-to/help-system-maintenance.md
@@ -170,7 +168,6 @@ features:
     description: Dynamic teams, workflow composition, and agent models
     files:
       - src/attune/orchestration/**
-      - src/attune/coordination/**
     tags: [orchestration, teams]
     doc_paths:
       - docs/how-to/multi-agent-coordination.md
--- a/.help/features.yaml
+++ b/.help/features.yaml
@@ -97,7 +97,6 @@ features:
     files:
       - src/attune/workflows/test_runner.py
       - src/attune/workflows/test_maintenance.py
-      - src/attune/workflows/test_lifecycle.py
     tags: [tests, debugging, fixes]
 
   mcp-server:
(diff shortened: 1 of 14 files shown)


silversurfer562:
Follow-up: the one dead symbol the guard surfaced (`WorkflowExecutionError` in `code-quality/error.md`) is now **fixed inline** in 9602a709d — corrected to the real `SdkSubprocessError` and the `_KNOWN_DEAD_SYMBOL_DRIFT` allow-list emptied, so the template-symbol check is strict for every feature with zero allowances.

PR #1057, opened 2026-06-24
fix(workflows): probe-fallback for SDK failure capture when argv is unrecoverable

silversurfer562:
## Problem

Dogfooding the ops dashboard (2026-06-06): every SDK-workflow failure showed
the user `(capture-call also failed: IndexError: list index out of range)`
instead of the real `claude` CLI stderr — defeating the
`sdk-error-message-fidelity` feature (marked complete in v7.3.0/7.3.1 but
broken on the current SDK).

## Root cause (verified against claude-agent-sdk 0.1.63)

The exception raised at `claude_agent_sdk/_internal/query.py`
(`raise Exception(message.get("error", "Unknown error"))`) carries **no argv**:
`args[0]` is the string `"Command failed with exit code 1"`, `__cause__` is
`None`, and there is no `.cmd`. So `_last_subprocess_argv` correctly returns
`[]`, and `subprocess.run([])` then raised the `IndexError`.

The task 's suggested fix #1 (extend the extractor to match the SDK shape) has
nothing to match — the argv is genuinely unrecoverable. The fix is #2: a
fallback probe.

## Changes

- **`capture_subprocess_failure`**: when `args` is empty, fall back to a minimal
  `claude` probe instead of `subprocess.run([])`. Added `stdin=DEVNULL` so the
  probe can't hang on inherited stdin. Honest `"claude CLI not found"` message
  when no binary exists. One source edit fixes all 17 call sites unchanged.
- **`_fallback_probe_argv()` + `_find_claude_cli()`** (new): build
  `[<claude>, "-p", "ping"]`, locating the binary in the SDK's own order
  (bundled → PATH → known locations) so the probe reproduces the same
  auth/quota/not-found failure the workflow hit.
- Rewrote `_last_subprocess_argv` docstring to document the verified 0.1.63
  reality.

## Tests

- `test_real_sdk_exception_shape_yields_empty`: drift guard pinned to the
  installed SDK — reproduces the real exception, asserts `[]`, and its failure
  message tells a future maintainer to teach the extractor the new shape if the
  SDK ever starts stashing argv.
- Replaced the old empty-argv test with three: no-IndexError guarantee,
  probe-runs-and-classifies (`api_quota` — the exact dogfood scenario), and
  honest not-found message.
- New `TestFallbackProbe` class for the probe/finder helpers.

## Verification

- 32/32 in `test_sdk_error_fidelity.py`, 18/18 across phase 2/4/5/6 regression
  files, ruff clean.
- End-to-end: the dogfood "usage limits" stderr now classifies as `api_quota`
  instead of the `IndexError` string.

### DIFF
--- a/src/attune/workflows/agent_sdk_adapter.py
+++ b/src/attune/workflows/agent_sdk_adapter.py
@@ -11,6 +11,7 @@ from __future__ import annotations
 
 import logging
 import os
+import platform
 import re
 import shutil
 import subprocess
@@ -679,7 +680,22 @@ def capture_subprocess_failure(
         itself raises. The synthetic strings are intentionally
         formatted so the classifier's "unknown" fallback still
         renders them to the user.
+
+    Note:
+        An empty ``args`` means ``_last_subprocess_argv()`` could not
+        recover the failing argv from the SDK exception (the common
+        case on current SDK versions — see that function's docstring).
+        Rather than calling ``subprocess.run([])`` (which raises
+        ``IndexError`` and surfaced the useless "(capture-call also
+        failed: IndexError ...)" message), fall back to a minimal
+        ``claude`` probe that reproduces the auth/quota/not-found
+        failure modes we care about.
     """
+    if not args:
+        args = _fallback_probe_argv()
+        if not args:
+            # No claude binary locatable — honest synthetic message.
+            return "(capture-call also failed: claude CLI not found)"
     try:
         result = subprocess.run(  # noqa: S603
             args,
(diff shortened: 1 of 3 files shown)


silversurfer562:
Superseded by main. The probe-fallback this PR introduced (`_fallback_probe_argv`) was independently solved on main, more completely:

- **Probe-fallback when argv unrecoverable** → main's `_claude_health_probe_argv()`
- **Keep the probe out of unit tests** → main's `_sdk_error_probe_enabled()` gate (default-OFF via `ATTUNE_SDK_ERROR_PROBE`), cleaner than this PR's autouse-fixture approach
- **Tests** → main has a superset (4 vs 2)

After resolving the conflicts toward main, the only surviving change was a `conftest.py` fixture that patches `_find_claude_cli` — which no longer exists on main, so it would `AttributeError` across the workflow unit tests. Nothing unique left to salvage. Closing.

Issue #1568, opened 2026-07-21
Resume ops-dashboard-polish remainder — B2, C1, C3, D-sweeps


silversurfer562:
**Spec:** docs/specs/ops-dashboard-polish (parked 2026-07-13, Resume-Trigger: evergreen) · approved task table in tasks.md.

Remaining: B2 run-view disk-fallback for evicted runs; C1 `/memory` read-only page; C3 memory nav + KPI counters; D1 tooltip sweep + CI grep gate; D2 empty/error-state sweep; D3 keyboard-nav audit; D4 contrast audit; D6 visual consistency sweep.

Each is a small independent PR. Feature code → **held draft** during the 10.6.0 hold; test/docs-class items may merge immediately.

State derives from the spec files — verify tasks.md before building; close when the table is done.

silversurfer562:
2026-07-21 autonomous pass — every remaining row is now built or resolved:

| Item | Outcome |
|---|---|
| B2 | already done (row was stale; verified in the morning session) |
| C1 `/memory` | **premise re-validated** (targets the post-#1239 Redis-derived index, not `~/.attune/memory/`) → built in held draft #1576, with a chair flag: #1545 ruled Sessions off the same day — approve or rule off at the sitting |
| C3 memory half | rides in #1576 (nav + Home KPI that hides on unreachable Redis) |
| C2 | status truth: shipped then **deliberately removed in #1545** — row updated |
| D1 | already in held draft #1571 |
| D2 | sweep executed as a regression lock (`test_empty_states.py`, 9 green, zero production changes needed) → #1577, auto-merge armed |
| D3 | audited; one real defect (editable pill suppressed its keyboard focus ring) fixed in held draft #1578 |
| D4 | audited with measured WCAG ratios; `.chip-muted` light-mode hard fail (2.31) fixed in #1578; 0.11 residual to strict AA flagged for the chair |
| D6 | audited; tables/chips single-sourced; dual empty-state idiom ruled a convention (decisions.md) |

Close this issue when the held queue lifts and #1576's chair call is made.

silversurfer562:
Resolved 2026-07-28 — both close conditions from the 2026-07-21 pass are now met.

That comment said: *"Close this issue when the held queue lifts and #1576's chair call is made."*

**1. The held queue lifted** — every held draft named in the table is merged:

| PR | Row | State |
|---|---|---|
| #1571 | D1 tooltip unification | MERGED |
[...]

PR #1576, opened 2026-07-21
feat(ops): /memory page over the Redis-derived index — C1+C3

silversurfer562:
## ops-dashboard-polish C1 + C3 memory half (#1568)

**Held draft — merges at the 2026-07-27 sitting** (`hold-until-07-27`).

C1's premise was re-validated before building (recorded in the spec's decisions.md): the original row targeted `~/.attune/memory/`, which memory-unification #1239 retired. This page reads the REAL serving layer — the `attune:memory:*` derived index in local Redis (~1,050 keys live: 766 lessons, 166 file pointers, 93 edges).

- **`attune.ops.memory_data`** (framework-free): family counts, paginated node table (50/page), type-dispatching node detail. Every read returns `None` on unreachable Redis → the page renders an explanatory empty state, never a 500. Detail view refuses keys outside the `attune:memory:` namespace.
- **Routes/templates**: `GET /memory`, `GET /memory/node` — `data-tooltip` (never `title=`, per D1), specs-style table + empty-state classes.
- **C3 memory half**: `/memory` in top nav; Home **Memory nodes** KPI that hides when Redis is down instead of showing a lying zero.
- **Dogfood catch → regression test**: plain-string keys in the namespace (`attune:memory:context`) WRONGTYPE'd the HMGET pipeline and degraded the whole page on first live render; fixed with `execute(raise_on_error=False)` + type dispatch, locked in `tests/unit/ops/test_memory_page.py`.

**Receipts**: 17 unit tests green (data layer + routes, faked Redis client); live browser verification against the real index — family tiles, table, node detail, Home KPI (screenshot in session); nav renders on every page.


silversurfer562:
**Chair flag for the 07-27 review:** this page was built the same day #1545 chair-ruled the Sessions page OFF the dashboard. The rationales differ — Patterns was structurally empty and Sessions was ruled off after the ops research pass, while `/memory` reads ~1,050 live keys and serves the memory-is-the-product thesis — but the same lens could reasonably rule a memory browse page off too. This draft is held so that call is deliberate: approve → `gh pr ready` + merge at the sitting; rule it off → close unmerged, and the C1 row gets a `removed-by-ruling` status like C2's. Recorded in the spec's decisions.md.

PR #1630, opened 2026-07-23
fix(scripts): fail closed when capability projector imports a foreign checkout

silversurfer562:
## Summary

Found 2026-07-23 during a 14-PR assembled-tree preflight: running
`python scripts/project_capabilities.py` puts `scripts/` at
`sys.path[0]` — not the cwd — so `import attune_redis` (and
`attune` itself) falls through to the editable install and imports
the MAIN checkout's packages. The server then registers the wrong
plugin tool set and the projector silently derived 55 instead of the
tree's true 60 (the 5 `session_memory_*` tools from #1594 were
missing). The existing fail-closed subset guard (core ⊆ registered)
cannot catch this because it still holds either way.

## Fix

After building the server, `derive_values()` asserts every imported
in-repo provider package (`attune`, `attune_redis`) resolves under
the repo root, raising `ProjectionError` (fail closed, never write)
with the `PYTHONPATH` remedy otherwise.

## Receipts

- 45/45 tests pass in `tests/unit/scripts/test_project_capabilities.py`
  (4 new: foreign-checkout fire, clean-env pass, absent-package skip,
  unlocatable module).
- Live repro from the worktree: bare `python scripts/project_capabilities.py --check`
  now exits 2 naming the foreign resolution + remedy; with
  `PYTHONPATH=<wt>:<wt>/src` it derives from the worktree and passes.

### DIFF
--- a/scripts/project_capabilities.py
+++ b/scripts/project_capabilities.py
@@ -73,6 +73,41 @@ class ProjectionError(RuntimeError):
 
 # --- Derivation --------------------------------------------------------
 
+# Provider packages that live in this repository. If one is importable
+# it must resolve to THIS checkout, or the derived counts describe some
+# other revision's tool set.
+_IN_REPO_PACKAGES = (("attune", "src/attune"), ("attune_redis", "attune_redis"))
+
+
+def _assert_in_repo_imports(repo: Path) -> None:
+    """Imported in-repo provider packages must resolve to this checkout.
+
+    Running the script as a file puts ``scripts/`` (not the repo root)
+    at ``sys.path[0]``, so ``import attune_redis`` can silently fall
+    through to another checkout via the editable install and register a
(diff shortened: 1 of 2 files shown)


silversurfer562:
Second commit `992a68054` reconciles with the originating session's two-layer plan: **layer 1 self-heal** — `derive_values()` now prepends `repo/src` and `repo` to the front of `sys.path` before importing, so a bare worktree run derives the tree's own counts (verified live: exit 0 with the worktree's 25/53/47, no PYTHONPATH needed) — and **explicit `attune_redis` import** so the provenance guard never depends on the server's best-effort plugin hook. Guard from the first commit stays as the fail-closed backstop. 47/47 tests. Derivation semantics unchanged (07-27 step-3b expectations unaffected).

PR #1846, opened 2026-08-01
docs(lessons): EOD batch 7 — import-side-effect registries emit false signal on cold miss

silversurfer562:
One from the morning sweep: a side-effect-populated registry needs an ensure-loaded call at its read seam, cold paths need subprocess tests, and a miss path that emits telemetry converts a loading bug into corrupted signal (#1845's cold-import fix).

### DIFF
--- a/.claude/lessons.md
+++ b/.claude/lessons.md
@@ -20575,3 +20575,26 @@ def ", start_idx + 1)` for module-
   finding), not how the system behaves — verify against the tree
   before writing machinery, and record the dissolved concern as
   design text so the next cold reader doesn't re-raise it.
+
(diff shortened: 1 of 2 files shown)


silversurfer562:
Windows 3.12 failure diagnosed: `test_slow_sink_does_not_block_sources` asserts a wall-clock bound (`elapsed < 1.0`) and a loaded runner measured 1.578s with zero actual blocking — the diff under test is lessons-only. Fixed the flake at its root in `566c1e289`: the bound is a regime check (inline-await ≈10s vs fire-and-forget), so it's now 5.0s with 2x margin both ways, and parked sink tasks are released after measurement. PR stays in the tests+docs auto-merge class; label re-evaluates on the new head.

PR #1971, opened 2026-08-06
docs(specs): approve broad-except-ratchet; rule Q1-Q3

silversurfer562:
Chair approved the broad-except-ratchet spec on 2026-08-06 via the three-question form. Docs-only: a status flip plus the ruling record.

## Rulings

| Q | Ruling |
|---|---|
| Q1 — `# noqa: BLE001` sites | **Counted.** The annotation is not an exemption. |
| Q2 — scope | **`src/attune` + `attune_redis` + `backend`.** |
| Q3 — fires-on-violation receipt | **At implementation**, in the build PR. |

## The measurement that drove Q1

| Tree | Sites | Files | Of which annotated |
|---|---|---|---|
| `src/attune` | 586 | 247 | 580 |
| `attune_redis` | 31 | 7 | 30 |
| `backend` | 7 | 5 | 3 |

Q1 offered "treat the annotation as the documented-contract marker" as a plausible reading. It isn't tenable: **580 of 586** sites carry it. The annotation records that ruff was satisfied, not that a contract was documented. Excluding annotated sites would have left the ratchet policing **6 sites out of 586** — a guard in name only.

## On Q2

The chair picked the widest scope and explicitly asked the lead for judgment and pushback. The lead has **none to offer**, and says so rather than manufacturing a counter-case: `backend/` is 7 sites (cheaper than one busy `src/attune` module) and holds auth and subscription code where a swallowed exception is the worst instance of the class; `attune_redis` ships bundled in the wheel, so a directory boundary shouldn't shelter shipped code. One caveat recorded, not argued: `backend/`'s last substantive commit is 2026-06-21, so if that tree is later retired its baseline entries retire with it — the ratchet is not a reason to keep it alive.

Separated from #1970 deliberately: spec status flips are merge-now (docs-outbox R2) and shouldn't be gated behind a feature PR's CI.

### DIFF
--- a/docs/specs/broad-except-ratchet/decisions.md
+++ b/docs/specs/broad-except-ratchet/decisions.md
@@ -1,0 +1,77 @@
+# broad-except-ratchet — decisions
+
+## D1 — APPROVED; the three open questions ruled
+
+**Date:** 2026-08-06 · **Status:** approved (chair: Patrick, via the
+three-question approval form; Q2 answered with an explicit request
+for the lead's judgment and pushback)
+
+| Q | Ruling |
+|---|---|
+| Q1 — `# noqa: BLE001` sites | **Counted.** The annotation is not an exemption. |
+| Q2 — scope | **`src/attune` + `attune_redis` + `backend`.** |
+| Q3 — fires-on-violation receipt | **At implementation** — seeded and proven in the build PR. |
+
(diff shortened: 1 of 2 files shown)


silversurfer562:
Superseded by #1972 (identical content, renamed branch to escape a wedged Tests workflow). Merged as 59f8cc7ce.

PR #1987, opened 2026-08-08
feat(memory): redis_health_check doctor diagnostic (redis-config-truth rct-3)

silversurfer562:
## Summary

rct-3 of the redis-config-truth ladder (R2): `redis_health_check` now carries the **doctor diagnostic** — a redacted effective-config section derived from the canonical resolver, so users can see exactly which env vars their connection came from and why auth is failing, without secrets ever appearing.

- New `_effective_config_report()`: resolver `source_map` (which var supplied url/password/user), `redacted_url`, recorded `overrides`, and the R3-classified `health` state + scrubbed detail.
- Handler payload extended additively (`effective_config`, `backend_selected`); the tool's input_schema stays frozen (D6).
- Drift rule honored: the report is derived FROM `resolve_redis_connection()` + `classify_redis_health()` — no independent env reads.
- Degrades gracefully when attune core isn't importable; the health tool's fail-open posture is unchanged.

## Receipts

- 9 new tests (`tests/unit/memory/test_redis_doctor_diagnostic.py`): incident shape names `REDIS_URL` (password-less) + `REDIS_PASSWORD` (present-and-merged); whole-report JSON grep finds no secret; malformed config reports scrubbed detail without a source_map; core-absent degrade; handler integration.
- Ratchet sweep green locally (tests/unit/gates/ + tests/unit/quality/, 210 passed serially).
- Live-fire on the real requirepass env: `healthy`, sources named, password-leak grep False, `redis://:***@127.0.0.1:6379/0`.

tasks.md flip lands on this branch after #1986 merges (same status paragraph — avoiding a guaranteed conflict). D11 codex lane to follow before the merge read. Merge on explicit go only.


silversurfer562:
Codex D11 lane run (thread `review-claude-rct-3-doctor-diagnostic-20260808-1310`, advisory): **1 finding, real, fixed** in 846d6e4b4 — `classify_redis_health()` ran outside the helper's guarded block, so a classifier bug would have flipped an otherwise-healthy health-tool response to failure. Now guarded (broad-except baseline 12→13 with WHY comment); the degraded reason carries the exception TYPE only, so exception text can never leak config values. Regression test added; ledger row appended; ratchet sweep 220 green serially.

Remaining before merge read: tasks.md flip once #1986 lands. Merge on explicit go.

PR #1994, opened 2026-08-08
test(memory): self-provisioning requirepass regression lane (redis-config-truth rct-5)

silversurfer562:
## Summary

The redis-config-truth ladder's **final rung** (R4's regression guard): a non-mocked requirepass round trip that **provisions its own server** — the lane spawns an ephemeral `redis-server --requirepass` with a random password on a scratch port (auto-teardown), so the core incident AC is verified on every host with the binary instead of only machines that happen to run hardened Redis.

- Incident shape (password-less `REDIS_URL` + `REDIS_PASSWORD`) authenticates live through the resolver AND the rct-4 consumers (`connect_recall_redis`).
- Wrong password → live `degraded_auth` from a real NOAUTH/WRONGPASS (with the provisioned secret asserted absent from the notice detail).
- Missing password → live `AuthenticationError` — the pre-fix incident reproduced on demand.
- **Skip contract pinned by meta-tests**: skip ONLY on binary absence — a source-scan asserts exactly one `pytest.skip` guarded by `REDIS_SERVER_BIN is None` (self-count avoided via literal split), and a provisioning-must-work test fails loudly if the lane ever rots into skipping while the binary exists. A non-default-port assertion proves no reliance on a pre-configured instance.

No CI workflow changes: GH runners lack `redis-server`, so the lane legitimately skips there and runs on dev machines — where the incident class actually lives.

## Receipts

- 8/8 live-green against a genuinely provisioned server, serial AND under xdist (`-n 2`; `xdist_group` pins the module to one worker).
- Ratchet sweep green; status-line gate green on the tasks.md flip (rct-4 MERGED, ladder COMPLETE on this merge).

D11 codex lane to follow. Merge on the standing go ("merge the rct-5 PR when green").


silversurfer562:
Codex D11 lane run (thread `review-claude-rct-5-requirepass-lane-20260808-2342`, advisory): **3 findings, all real, all fixed** in 2417309d0 —

- [high] module-level `pytest.importorskip("redis")` was a second skip condition **evading the lane's own source-count meta-test**. redis is a core dependency — absence must FAIL, not skip. Now a plain import, and the meta-test additionally forbids the importorskip call form.
- [medium] the wrong-/missing-password live tests scrubbed only part of the env — all three live tests now share the full 11-var scrub helper.
- [medium] the kernel can legitimately assign 6379 as an ephemeral port, making the non-default-port independence proof flaky — `_free_port` now structurally refuses 6379.

Ledger row appended; 8/8 live-green serial + xdist. This is the ladder's final rung — standing go ("merge the rct-5 PR when green") applies; will label on full-matrix green.

PR #2021, opened 2026-08-10
feat(memory): open P3 + tasks 2 and 4 — curated serve telemetry and windowed reader

silversurfer562:
Opens **memory-status-integrity P3** on the chair's direction and ships its first two buildable rungs.

## P3 opened — the measure-first gate executed AT authoring (D8)

The phase's own text gates it: "needs recall telemetry that may not exist yet; measure before building." Measured against the live stream before the ladder was written: the pipe is healthy (4,217 events) but **every per-item identity is raw-tier** (`finding_ids` / `item_id` UUIDs) — no event names a curated stem, so R6's "how often the memory is actually served" had no data source. Recorded as decisions **D8**, with the D1 boundary pinned: frequency may reorder the REVIEW queue, never filter a recall surface. The 8-task ladder is ordered instrument → accrue → rank; OQ3 (weekly cadence) is pulled in as task 7.

## Task 2 — curated serve telemetry (the emitter)

`attune.memory.serve_telemetry.log_curated_recall`: a `curated_recall` event naming served stems, written to the SAME sink and envelope as the hook-side writer so the reader parses one stream. `ATTUNE_MEMORY_TELEMETRY` + `DO_NOT_TRACK` honored (identical semantics to the hook — one setting governs both writers). **Stems only, never content** — pinned by a test asserting the record carries the envelope fields and stems and nothing else. Fail-open by contract (ratchet-baselined with the WHY); path-validated write. Wired into `PersonalMemory.query()` and `recall_digest` fetches via `except ImportError`-only guards (the emitter never raises).

## Task 4 — the windowed reader

`serve_telemetry.serve_counts(window_days=30)`: per-stem counts from the sink **including rotated siblings** (a 30-day window can span the writer's size-rotation). Malformed lines/timestamps skipped, missing sink → all-zero, an unreadable sibling warns-and-skips without costing the live counts, glob errors fail open. Window/path/today injectable — fully hermetic tests.

## Receipts

- 249 passed serially (gates+quality + the 21-test serve_telemetry suite: record shape, stems-only pin, both off-switches, cap/truncation, unwritable-home, window/rotation/malformed-line matrix, both surface wirings).
- `serve_telemetry.py` **100% measured before codecov sees it** (both halves).

## Governance

Memory code → **D11 lane owed**; running the codex lane now. NOT auto-merge-armed; chair merge read after.

From merge, every curated recall starts accruing frequency data — task 8's receipt window opens. Remaining P3: 3 (hydrate emitter — external), 5 (the rank fold), 6 (acceptance regression), 7 (weekly cadence), 8 (receipt).


silversurfer562:
D11 lane RUN (codex, 8 sent / 0 omitted): **3 findings, all real — accepted and fixed in-branch** (`3b436b0b8`): (1) valid-but-non-object JSON lines crashed the reader's documented fail-open path — isinstance guard + 3-shape regression; (2) the window gained an upper bound so future-dated events can't corrupt serve frequency; (3) the lower bound is now exclusive — window=30 spans exactly 30 calendar dates, pinned at the boundary. Complexity ratchet respected via a per-line parse helper. 251 tests green serially; module still 100% measured. Ledger row appended. D11 precondition satisfied — awaiting chair merge read.

PR #2045, opened 2026-08-10
feat(memory): refs-v2 extraction + binder behind ATTUNE_MEMORY_REFS_V2 — mcv T1+T2 (chair-read)

silversurfer562:
## What

T1+T2 of the merged design (#2043), everything behind `ATTUNE_MEMORY_REFS_V2` **default OFF** — flag off is byte-identical v1 (pinned by test):

- **T1**: v2 prompt requests 0–3 `kind:value` refs (`file|pr|spec`) and stops requesting `source_ref`; parser extends `_parse_typed_findings` SYNTAX-ONLY under the R5#2 item-drop discipline — an unknown kind survives to the binder, an over-eager ref never costs a good finding.
- **T2**: binder in the hook — derives the tool_use universe (file-path keys + command-text patterns), binds by EXACT membership (lexical normalization only, no filesystem access), reason-coded rejections (`bad_kind`/`not_in_session`), honest `no_ref_universe` degradation (refs kept as proposed, never claimed rejected). Output rides the existing tag channel with `schema_version:2`.
- Probe script untouched by design: its D8 baseline arm must stay reproducible; T3's v2 arm imports the hook's functions.

## Receipts

- 25 new tests incl. the no-fuzzy drift guard, prefix-containment negative, universe-pollution regression, and flag-off-never-binds; serial `tests/unit/hooks/ + tests/unit/memory/`: **2897 passed, 49 skipped, 3 xfailed**.
- Gates suite 223 passed; pinned black/ruff clean.

## D11b lane (ran BEFORE this read)

Codex on the full diff: **3 findings, all real, all adopted** — universe pollution from non-command inputs (high), missed `gh pr view N` forms, dot-prefixed/root paths underivable. Ledger row on this branch.

## D11 pre-flight (self-run)

Fail-open: unreadable transcript → `no_ref_universe`, binder isolated in try at the call site. Windows: no encoding/subprocess surface; paths normalized lexically (`os.path.normpath`). Record-before-stamp: sentinel written after stash (unchanged). Path containment: exact membership, prefix-containment negative test. Key collisions: universe keyed by kind. USERPROFILE: no home redirection.

Windows lanes should be green before merge (path normalization is lexical but the matrix is the receipt).


silversurfer562:
Windows lanes fixed in bd34df4e7: `_normalize_ref_value` used `os.path.normpath`, which emits backslashes on Windows — the `ref_bound:file:` tag became OS-dependent (`\repo\src\a.py`) and `test_binder_binds_exact_file_pr_and_spec` failed on both windows-latest hook lanes. Now wrapped in `Path(...).as_posix()` so the tag is platform-stable; binder and universe normalize through the same function, so binding semantics are unchanged. Verified locally: 25/25 refs-binding tests on POSIX + ntpath/PureWindowsPath simulation of both ref shapes. Pushed from the monitoring session (branch was level with origin; the session-2 worktree was not touched — it is now 1 behind, fast-forward).

PR #2046, opened 2026-08-10
docs(specs): D11 live-fire receipt probe (chair-read)

silversurfer562:
Live-fire probe of the chair-read gate shipped in #2044. This PR is docs-only — under pre-#2044 behavior the auto-labeler would apply `auto-merge-safe` within a minute. Expected NEW behavior: the label job logs the chair-read skip and never labels. Will be closed after observation; the same diff re-opens unmarked as the control.


silversurfer562:
Probe complete: label job executed the chair-read skip (`chair-read marker -> never auto-labeling`, 23:06:18Z, run 31441014984) and no auto-merge-safe label was applied. Re-opening the same diff unmarked as the control.

PR #2100, opened 2026-08-19
fix(agents): LLM response can no longer overwrite bandit severity counts in the Security gate

silversurfer562:
## Summary

`SecurityAuditorAgent._execute_tier` merged the LLM-enhancement response with a wholesale `findings.update(llm_findings)`, letting a hallucinated low/zero `critical_issues` count overwrite the bandit-parsed value **before** the gate decision (`critical == 0`). A bandit run with real critical/high findings could therefore pass the release Security gate. The same findings dict feeds the team-level `max_critical_issues` gate in `release_prep_team.py`, so both gate reads were exposed.

Found by the 12.0.0 post-release self-review (code-review workflow, 2026-08-19); verified by reading the source before fixing.

## Fix

- The severity counts parsed from bandit — `critical_issues`, `high_issues`, `medium_issues`, `low_issues`, `total_findings` — are now authoritative (`_BANDIT_AUTHORITATIVE_KEYS`). The LLM merge excludes them and may only enhance non-gate fields (`score`, `confidence`, `top_findings`).

## Tests

- New regression test `test_execute_tier_llm_cannot_overwrite_bandit_critical_count`: bandit-parsed `critical_issues > 0` + an LLM response claiming zeros must still fail the gate. **Verified to fail against the pre-fix code** (the LLM zero passed the gate) and pass with the fix.
- `test_llm_findings_merge_into_bandit_results` previously pinned the buggy overwrite behavior (asserted the LLM's count won); updated to the new contract.
- Full related suites green: 172 passed across `test_release_prep_team_orchestration.py`, `test_release_prep_team.py`, `test_release_agent_analysis_branches.py`, and `tests/unit/agents/release/` — the principle-7 sentinel semantics stay pinned.

## Note

Per D11, this is a lead-authored diff on a security/gate surface — flagging for the different-model review lane before chair read.


silversurfer562:
**Codex D11 cross-review lane completed** (thread `review-claude-quirky-morse-a941d6-20260819-1334`, manifest **3 sent / 0 omitted** — full diff coverage).

One HIGH finding, disposition **real — accepted with amendment** (ledger row in `docs/specs/cross-review/receipts.md`):

> Discarding all LLM severity counts prevents an LLM-detected critical vulnerability from failing the release gate when Bandit reports zero; preserve Bandit findings against downward overrides while still incorporating stricter LLM counts.

Lead triage: premise overstated (the LLM only re-analyzes bandit's own `stdout[:3000]`, so it can't detect what bandit never reported) — but a stricter *reclassification* (bandit MEDIUM judged critical) could indeed no longer fail the gate. Fixed in 558b3249c with a fail-closed **upward-only ratchet**: an LLM severity count may raise the bandit-parsed value, never lower it. Counter-case recorded: `_parse_response`'s regex fallback could yield a spurious nonzero count and block a clean release — chair-recoverable friction, preferred over the silent-bypass direction.

Receipts: new ratchet test + the zeros-cannot-lower regression test both green; 173 release-agent tests pass serially.

PR #2113, opened 2026-08-20
fix(cli): cost commands honor their exit-code contract on non-OSError failures

silversurfer562:
## Summary

Library-review R5 finding (checkpoint-1 probe pass 2026-08-20, `~/.attune/reports/attune-ai-review/`): the cost CLI handlers promise **"0 on success, 1 on failure"** but caught `OSError` only — corrupt tracker data (a `KeyError` on a record missing `timestamp`, a `TypeError` from `json.dumps` of unserializable data) crashed the CLI with a raw traceback.

## Changes

- `cmd_costs`, `cmd_costs_today`, `cmd_costs_export` each gain an `except Exception` fallback that logs via `logger.exception` and returns 1. The specific `OSError`/`PermissionError` branches and their messages are unchanged and still fire first.
- `cmd_costs_reset` is deliberately unchanged: it only performs file ops (`exists`/`unlink`), where `OSError` is the complete failure surface.
- Broad-except ratchet baseline gains `src/attune/cli_commands/cost_commands.py: 3` with the reason recorded at the entry — the baseline is the spec's sanctioned escape hatch for contract-required broad catches, and each call site carries a WHY comment.
- Four regression tests pin the behavior (KeyError via the workflow filter, TypeError via `json.dumps`, ValueError via a corrupt today-record, TypeError via `json.dump` on export).

## Receipts

- `tests/unit/test_cli_cost_commands.py` + `tests/unit/gates/test_broad_except_ratchet.py`: 40 passed serially.
- Pinned black + ruff pre-flighted clean before staging.


silversurfer562:
D11 cross-review lane complete (codex, 3 sent / 0 omitted — full coverage). One medium finding, triaged **real** and fixed in 5b6b8380c: `_export_json` now serializes via `json.dumps` before opening the destination, so a mid-serialization TypeError can no longer leave a truncated export or destroy a previously valid one. Regression test asserts no partial file remains. Ledger row appended to docs/specs/cross-review/receipts.md.

PR #2168, opened 2026-08-22
fix(security): close the three Low findings from the 13.0.2 review

silversurfer562:
## What

The three Low security findings from the 13.0.2 post-release self-review. Each was verified against the live tree before being fixed, and each is **mutation-checked** — reverting the fix fails the new test.

This touches `src/`, so it is **not** the Class-1 auto-merge lane. It is also a security diff, which under D11 is a risk class wanting a different-model review lane before the chair reads.

## 1. Constant-time client-token compare — `ops/security.py`

`x_attune_client != _SESSION_TOKEN` short-circuits on the first differing byte, so response latency leaks a matching prefix and the session token can be recovered byte by byte.

Now `secrets.compare_digest`, comparing **bytes**: `compare_digest` raises `TypeError` on non-ASCII `str`, and the header is caller-controlled — a `str` compare would turn a hostile header into a **500 instead of a 403**.

Verified: valid accepted; missing / empty / wrong / non-ASCII / lone surrogate / embedded NUL all → 403, none 500.

### One deliberate non-change, flagged for review

The `None`-token branch is **preserved**. The ops route tests' autouse fixture nulls `_SESSION_TOKEN` so header-less requests pass. Making absence deny — which principle #7 ("absence is not a pass") argues for — is defensible, but it is a **test-fixture change, not part of the timing fix**, and I measured it at **92 failing tests**. Left as-is with a comment naming the tradeoff, raised here rather than bundled. Happy to do it as a follow-up if you want it.

## 2. Option-like git ref — `patterns/git_extractor.py`

`git log -1 --format=X --help` prints usage instead of resolving a commit — demonstrated, not assumed. A ref starting with `-` is now refused before git runs, and a trailing `--` pins "no pathspec follows" so a ref that also names a file cannot be read as a path.

The guard is in `_get_commit_info` rather than at the call sites, so future callers inherit it. Today's only caller passes `HEAD~{i}`, so this is **defense-in-depth, not a live exploit**.

Chose a leading-dash guard over `--end-of-options` because the latter needs git >= 2.24, and for internally-controlled input a version floor is the worse trade.

## 3. Daemon umask — `monitoring/alerts_cli.py`

The review flagged this as *"real only if the daemon writes sensitive files with a bare `open()`"* and said it needed a second check. **That check was done.**

`AlertEngine` is constructed *before* `_daemonize()`, so the database and its directory are created under the normal umask — the report's worry, as literally stated, does not fire. But the watch loop reconnects afterward, and SQLite's journal/WAL siblings are created **inside** the daemon with no explicit mode.

Measured:

| umask | created directory | SQLite database |
|---|---|---|
| `0` (before) | **0o777** — world-writable | 0o644 |
| `0o077` (after) | 0o700 | 0o600 |

The in-code comment carries the **measured** values, not the estimate I first wrote (I had guessed 0o666 for the DB; it is 0o644).

## Tests

- **10** for the token compare — including a spy pinning that `compare_digest` is what actually runs, and that both sides are `bytes`
- **6** for the git ref — option-like refs refused *without invoking git*; normal refs still resolve; the trailing `--` asserted
- **2** for the umask — a static AST guard needing no fork, plus a POSIX-only behavioural check

The umask module computes its skip condition with a safe accessor, since a `skipif` **condition is evaluated at collection time** — a bare `os.geteuid()` would error the whole module on Windows.

## Verification

- `tests/unit/{monitoring,ops,patterns}/`: **2,165 passed, 1 skipped**
- `tests/unit/gates/`: **329 passed**
- Full `tests/` collection: **25,314**, zero import errors
- **Mutation**: each fix reverted individually fails its own test
- Pinned `black`, `ruff`, `ruff-bare-exception-check`, **`bandit`**: pass
- `check_badge_freshness.py`: OK

## Also found, not fixed here

`_daemonize()` does `os.chdir("/")` while `AlertEngine`'s `db_path` defaults to the **relative** `.attune/alerts.db`. After the chdir that resolves to `/.attune/alerts.db`, and `sqlite3.connect` fails with `unable to open database file` for a non-root user — so `attune alerts watch --daemon` looks broken. That is a correctness bug, not a security finding, so it is reported rather than folded in.


silversurfer562:
## Cross-review lane run (D11) — 1 finding, real, fixed

Chair requested the different-model review after this PR was already armed with `auto-merge-when-green`. The arm was **removed and native auto-merge disabled** before the lane ran, so the review could not be overtaken by its own subject. PR is held open.

**Seat:** codex (non-authoring). **Manifest:** 6 sent / 0 omitted — full diff, no truncation.

### Finding — `src/attune/ops/security.py:72` [medium] — CONFIRMED

> Encoding with `surrogateescape` still raises `UnicodeEncodeError` for caller-controlled lone surrogates outside U+DC80–U+DCFF (for example `\ud800`), so `require_client_token` can still surface a 500 instead of the promised 403.

Verified by execution:

| header | before |
|---|---|
| `\udcff` (U+DCFF — in range) | 403 ✅ |
| `\ud800` (U+D800 — out of range) | **UnicodeEncodeError → 500** ❌ |

`surrogateescape` only round-trips the surrogates it produces itself (U+DC80–U+DCFF). Fixed with `errors="replace"`, which cannot raise for any `str`.

### The part worth recording

**My own test masked it.** I parametrized exactly one surrogate and happened to pick `\udcff` — the top of the covered range — so the case passed and read as proof that hostile headers 403 rather than 500. The test was the reason I believed the claim in the first place.

Added `U+D800`, `U+DBFF`, and a surrogate embedded in ASCII. Mutation-checked: reverting to `surrogateescape` now fails 3 of them.

Same class as the 2026-08-20 tier-3 ledger row — an assertion that cannot distinguish the fixed code from the broken code.

### State

- Fix + strengthened tests: `5cc02b76f`
- R5 ledger row appended to `docs/specs/cross-review/receipts.md`
- 2,497 passed / 1 skipped across ops, monitoring, patterns, gates; pinned black + ruff clean
- **Not re-armed** — awaiting your read.

PR #2173, opened 2026-08-22
feat(classes): Phase 0 — tracked rule pack + derived register

silversurfer562:
## Summary
Release-audit-stage Phase 0 (R1+R2, spec merged in #2171; builds on #2172's package):

**Rule pack (R1)** — `attune.classes.rules` promotes the machine-local sweep suites (v1 R1–R6 + v2 R7a/R7b, library review 2026-08-20) into tracked code. Each rule carries its calibration receipt **verbatim**: R7a/R7b bind to attune-ai (recall 8/11, precision 1.0, PR #2121 ground truth) and are `calibrated-here`; R1–R6 carry only the attune-forms suite-level figure and run **advisory everywhere** — an uncalibrated rule never blocks and never clears. `attune.classes.scan` emits JSON with per-rule calibration state and advisory labeling; a crashing rule is recorded and fails the exit code (fail closed, contract §7).

**Derived register (R2)** — `attune.classes.register` computes status, never reads a hand-maintained column: CLOSED / BROKEN-GATE (loud) / FIXED-BUT-UNGATED / OPEN / DEFERRED / UNMECHANIZED. Gate mappings are identity-checked (stable node id + `Register-Class:` tag in the gate file; drift guard fails CI on rename or reassignment). DEFER records validate fail-closed with **release-count expiry via real git tags** — an expired DEFER's class reverts to OPEN, the block resumes. A dispositions file makes the review's dismissed-with-reason discipline machine-readable.

## Receipts
- Behavioral parity: tracked pack vs both machine-local suites on `src/attune/memory` — identical per-rule counts.
- First live derivation caught the hand register stale in **both directions**: H1/G1/G2/I-4 gates already existed (register still said fixed-but-ungated), and C3/C4a derive BROKEN-GATE because their gates assert a deliberately-scoped subset — the dispositions file is the fix; populating it from the review's triage ledgers is the named follow-up.
- 47 new tests (76 in the package); repo-identity and defer-expiry tests run through real git repos; broad-except in scan.py baselined per the ratchet's hatch with the WHY at the call site.

Also deletes the merged #2171 branch's handoff per its own instruction.

Spec: `docs/specs/release-audit-stage` · Next: populate dispositions, then Phase 1 (stage + packet).


silversurfer562:
## Windows lanes fixed — one test, not many

Pushed `20020c4e1` from another session at the chair's request. **The branch's own worktree (`roundtable-audit-workflow-b99ec3`) was not touched** — staged on a local branch off the fetched tip and pushed to this ref, so that session can `git pull` when it resumes.

### What was actually failing

**One test**, `test_dispositions_subtract_dismissed_hits`, on all 5 Windows lanes. The 6 red checks were that one test × 5 Python versions plus the `test-matrix-complete` sentinel it trips. macOS and Ubuntu were green throughout.

### Root cause

`_subtract_dispositions` compared a disposition's `path` verbatim against `str(Path(hit).resolve().relative_to(root))`. `str()` renders the **native** separator:

| | value |
|---|---|
| hit on Windows | `src\\mod.py` |
| disposition file (hand-authored) | `src/mod.py` |

The tuple never matched, so **every disposition was silently ignored on Windows** — dismissed hits still counted, and a class that should derive `FIXED-BUT-UNGATED` derived `OPEN`. Both sides now normalise with `.as_posix()`, which also lets a hand-written backslash path match on any platform.

### A note on the regression test

The behaviour **cannot be simulated off-Windows** — on POSIX a backslash is an ordinary filename character, so `Path` won't treat it as a separator and `as_posix()` leaves it alone. My first two regression tests tried to fake it and were wrong, not the fix. They're replaced with a source guard asserting both sides normalise, which runs everywhere, so a revert fails fast rather than waiting on a lane that isn't required to merge. Mutation-checked: restoring the old form fails the guard on macOS.

### Codecov is not failing

`codecov/patch` and `coverage` both **pass** — 85.37% patch against the repo's `target: 80%, threshold: 5%`. The ❌ in the Codecov comment is its own missing-lines commentary, not a gate.

### Verification

- `tests/unit/classes/` + `tests/unit/gates/`: **409 passed**
- Pinned `black` + `ruff`: clean
- Base was unmoved at push time (`79111b16f`)

Not armed for merge — leaving that to whoever owns this PR.

PR #2200, opened 2026-08-23
fix(agents): the LLM ratchet can no longer lift the Security gate's -1 sentinel

silversurfer562:
## What

#2194 made an unreadable bandit run return `critical_issues=-1` so the gate fails closed — but in `RELEASE_LLM_MODE=real` the severity ratchet (`if llm_count > findings[key]`) lifted `-1` to the LLM's `0`, and `success = critical == 0` passed the gate blind again. Reproduced: `success=True, critical_issues=0, note="Could not parse bandit output"`.

**Found by bug-predict** — a dogfood run on `security_agent.py` ranked this its top HIGH at the right line. Round table `q-bug-predict-health-001` (3/3 seats; chair-promoted C1).

## Fix

One extra condition in the ratchet: it applies only to a non-negative count, so the did-not-run sentinel is absorbing. The LLM can still raise a real count.

## Receipts

- **suite**: 72 passed serially across `test_release_agent_analysis_branches` (2 new), `test_specialized_agents`, `test_release_prep_team_orchestration`.
- **live-fire**: the exact reproduction probe now yields `success=False, critical_issues=-1`.

## Governance

Security/enforcement surface → D11 lane owed before the chair reads.

### DIFF
--- a/src/attune/agents/release/security_agent.py
+++ b/src/attune/agents/release/security_agent.py
@@ -113,12 +113,18 @@ class SecurityAuditorAgent(ReleaseAgent):
                             }
                         )
                         # Fail-closed ratchet: a stricter LLM severity
-                        # count may raise the bandit value, never lower it.
+                        # count may raise the bandit value, never lower it
+                        # — and never REPLACE the -1 did-not-run sentinel:
+                        # 0 > -1, so without the guard an LLM reporting
+                        # "no issues" turned an unreadable bandit run back
+                        # into a clean gate (found by bug-predict dogfood,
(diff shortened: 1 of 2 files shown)


silversurfer562:
Superseded by #2203 (squash 7731579b7), which was stacked on this branch and carried commit 705b1c3cf — the sentinel guard (security_agent.py:194 on main) and test_llm_ratchet_never_lifts_the_did_not_run_sentinel are both on main. This PR went DIRTY against its own already-merged content, so closing rather than resolving.

PR #2250, opened 2026-08-24
feat(test-gen): write real test files — Write tool, validated output_dir, truthful result keys (#2213)

### LINKED ISSUE #2213: test-gen emits no runnable tests (no Write tool) — 'working' verdict is exit-0 only
## Summary

The `test-gen` workflow is marked "working" in the fleet dashboard, but
it emits **no runnable test code** — only prose describing tests. Its
green verdict is true only in the "exited 0" sense; it does not produce
tests a user can run.

## Evidence

Found by the planted-defect probe runner (`scripts/workflow_probe_runner.py`,
added in #2210), run live 2026-08-23 against
`tests/fixtures/workflow_probes/testgen/orders.py` (a small branchy,
untested module):

```
[FAIL] test-gen: workflow emitted no runnable test code
       (the report contained no pytest-shaped code fence)
       $0.62, 161s, emitted_test_chars=0
```

The workflow ran cleanly (no crash, real spend) and produced a report,
but the report contained no extractable pytest-shaped code fence.

## Root cause (structural)

`test-gen`'s subagents are configured with `tools=["Read","Glob","Grep"]`
and the workflow's `allowed_tools=["Read","Glob","Grep","Agent"]` — **no
Write/Edit**. So it can describe tests but cannot emit executable ones.
The prompt even asks "how many test files were written", which is
unbackable given the toolset. The MCP handler surfaces `tests_generated`
/ `output_path` keys the workflow never produces.

(See `src/attune/workflows/test_gen/workflow.py`. The variant that DOES
write files — `ParallelTestGenerationWorkflow` / `test-gen-parallel` —
is not in the dashboard registry and needs pre-existing coverage data.)

## Proposed fix (either)

1. Wire a Write tool + output path into `test-gen` so it emits runnable
   test files (and populate `tests_generated` / `output_path`), or
2. Relabel the verdict / dashboard UX so users don't expect runnable
   output from `test-gen` and are routed to `test-gen-parallel` when
   they want files.

## Context

Part of the workflow fleet-health validation (roundtable
`q-workflow-fleet-health-001`, chair-promoted into
`docs/specs/test-quality-program/`). Full write-up in that spec's
`decisions.md` (PRs #2210, #2211, #2212).

silversurfer562:
Closes #2213.

test-gen could only *describe* tests (no Write tool anywhere in its toolset), while the MCP handler surfaced `tests_generated`/`output_path` keys the workflow never produced — the "working" verdict was exit-0 only.

- **Write pipeline**: test-writer subagent + orchestrator get `Write`; `permission_mode="acceptEdits"` (the fix_workflow precedent); a Write-scope PreToolUse guard denies writes outside `output_dir`. `make_edit_scope_guard` moved to `agent_sdk_adapter` as the shared home (fix_workflow re-imports; its import path still works).
- **Validated `output_dir`** input: defaults to `<path>/tests/generated`; explicit values must resolve inside the target's directory (traversal rejected), enforced both at intake and at tool-call time.
- **Truthful result keys**: `generated_files` / `tests_generated` / `output_path` counted **from disk** (pre-run snapshot diff) — never from the model's claims.
- **Probe upgrade**: the planted-defect probe now executes the REAL files written under `tests/generated` (fence extraction kept as regression fallback), with `files_written`/`meta_tests_generated` evidence.

Receipts: 212 test_gen/fix/MCP-handler tests + 2,973 workflows tests green. Live-fire probe run in flight — result posted as a comment when it lands.

### DIFF
--- a/docs/specs/workflow-behavioral-validation/records/20260824T043208-test-gen.json
+++ b/docs/specs/workflow-behavioral-validation/records/20260824T043208-test-gen.json
@@ -1,0 +1,18 @@
+{
+  "workflow": "test-gen",
+  "fixture": "tests/fixtures/workflow_probes/",
+  "receipt_type": "executed-tests",
+  "verdict": "pass",
+  "cost_usd": 0.6521,
+  "duration_s": 160.2,
+  "ran_at": "2026-08-24T04:32:08Z",
+  "runner_version": "0.3.0",
+  "git_sha": "1e70fdb96",
+  "evidence": {
+    "emitted_test_chars": 3160,
+    "files_written": 1,
+    "meta_tests_generated": 1,
+    "pytest_tail": ".......................                                                  [100%]\n23 passed in 0.02s",
+    "reason": "emitted tests (files_on_disk) imported, ran, and passed"
+  }
+}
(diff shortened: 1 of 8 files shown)


silversurfer562:
Live-fire receipt landed: `[PASS] test-gen: emitted tests (files_on_disk) imported, ran, and passed` — $0.65, the probe took the real-files path (not the fence fallback), and the run record + registry re-projection are committed on the branch. Arming when-green.

PR #2271, opened 2026-08-24
feat(classes): mechanize C10 phantom read — R8/R9 repo rules + gated baseline

silversurfer562:
## What

Mechanizes class **C10 (phantom read)** — Register ✅ → **Mechanize** →
Calibrate → Gate, per the C10 register entry (2026-08-24) and the
repo-side record in `docs/specs/release-audit-stage/decisions.md`
(#2269). The two narrow rules the entry ruled calibratable; the general
cross-function dataflow form stays declined as uncalibratable.

## The rules (`src/attune/classes/rules.py`, new `REPO_RULES` pack)

Repo-level rules — the invariant spans producer and consumer files, so
they take a repo root rather than one file's tree (`RepoRule`, same
metadata shape as `Rule`, calibration receipts included verbatim).

**R8 — `scan_result_key_contract`** (result-key contract, MCP seam):
an MCP `_workflow_response` field pick must read a key its workflow can
emit — plain-dict `final_output` keys (legacy path), or, on
adapter-built (`from_agent_output`) producers, the findings-/score-like
names `_report_fields` serves from the parsed report. Any other pick
can only ever return its static default while the response renders
healthy. Producers with neither form are unresolvable and never judged
(precision over recall).

**R9 — `scan_entry_point_channels`** (channel identity, both
directions): every entry-point group `src/` reads must be named in
pyproject (section or the deliberate third-party comment) or carry an
explicit legacy marker (`LEGACY` binding name, or `legacy=True` at the
call); every group pyproject registers must be read by `src/` or be
externally consumed (`pytest11`). An empty read is a legal state, so
nothing else distinguishes "no producers" from "wrong channel".

## Calibration (probe-first, every hit hand-triaged)

| Rule | Surface examined | Hits | Real | False positives | Recall probe |
|---|---|---|---|---|---|
| R8 | 22 picks across `workflow_handlers.py` + `server.py` | 9 | 9 | 0 | fires on registered #2213 instance at `ac47cfb21^` |
| R9 | 8 group reads + 4 registered groups | 0 | — | 0 | 3/3 facets of the #2259 instance fire at `c7c94f33e^` |

R8's 9 current-tree hits are **real live instances** (basis: traced
both `_workflow_response` branches — `_report_fields` resolves only
findings-/score-like picks, `_legacy_fields` needs a plain dict, and
all five flagged producers are purely adapter-built; inferred from
code reading, not executed live):

- `_run_test_generation: tests_generated, output_path` — the #2213
  residual: #2250 made `result.metadata` truthful for probes, but the
  MCP pick surface cannot see metadata, so these fields always carry
  0/None.
- `_run_doc_gen: document, sections` · `_run_code_review: feedback` ·
  `_run_research_synthesis: key_insights, confidence` ·
  `_run_release_notes: approved, recommendation` — same shape.

**Class-level recall: 2/7 registered instances by design.** The
workflow-internal reads (#2222, #2223) and the semantic probe misreads
(severity-vs-category, count-as-gate, `success` semantics) sit outside
both seams and would need the declined dataflow rule.

## The gate (`tests/unit/gates/test_phantom_read_gate.py`)

`Register-Class: C10` tag + `GateRef` in `register.py` `GATES` (the
drift guard enforces the pair). `test_no_phantom_reads`: R9 asserts
zero; R8 asserts nothing outside a **shrink-only baseline** of the 9
calibration-time sites — mechanization precedes sweep-fix in the class
pipeline, and the baseline is the sweep-fix worklist. New phantom
picks are blocked outright; the stale-entry check carries the
scanner-blindness caveat from the 2026-08-21 allowlist-ratchet lesson.
Every calibration discriminator is pinned as a passing fixture
(dict-literal servable picks, findings-/score-like picks on adapter
producers — 13 of the 22 picks examined would be false positives
without it — unresolvable producers, control kwargs, all four R9
marker forms, non-group literals).

## Receipts

- Full tree: `pytest tests` — **25147 passed, 257 skipped, 6 xfailed, 0 failed** (serial gate/classes runs also green).
- Coverage on `attune.classes.rules`: **95%** (worktree-safe measurement; remaining misses are OSError/parse-degradation lines).
- Complexity ratchet: refactored the three flagged blocks below grade D (no baseline additions).
- Brand-drift ratchet: legacy group names removed from shipped docstrings (fixtures keep them; test dirs are outside the guard's contract).

Next pipeline step after merge: sweep-fix the 9 baselined sites, then
close C10. Machine-local register (`~/.attune/reports/attune-ai-review/CLASS-REGISTER.md`)
updated to MECHANIZED + GATED with this PR.


silversurfer562:
## Post-open additions (2 commits)

**Codex D11 lane** (run before the chair's read; ledger row appended to `docs/specs/cross-review/receipts.md`): 3 findings, 3 real.

- **Fixed** (high): R8 discovered only handler-nested workflow imports — a module-level import silently evaded the seam. Top-level imports now count for every handler in the file, collected from **direct body statements only** (the lead corrected the implied remedy: walking the whole tree would have unioned every handler's nested imports into every other's contract). Fixture pins both directions.
- **Fixed** (high): R9's file prefilter demanded the plural `entry_points` substring while its collector matches singular helper names like `_load_entry_point_workflows`. Relaxed to the singular form; fixture pins a helper-only file.
- **Accepted, no code change** (high): `_producer_contract` unions final_output dict keys module/package-wide, so an unrelated producer site can make a phantom key look servable. Real as a *recall* limitation and consistent with the rule's calibrated precision-over-recall posture — the union can only make R8 quieter, never noisier.

Calibration re-verified post-fix: current tree 9 R8 / 0 R9 hits (identical identities); pre-fix trees still fire (9 with #2213, 3/3 entry-point facets).

**Retro O3 fold-in** (chair-approved, relayed from the good-morning session): `REGISTERED_CLASSES` manifest in `register.py` — `{class_id, registered_date, register_ref}` unioned into the derived universe, so a REGISTERED-but-unmechanized class is visible to `attune classes register` as UNMECHANIZED with a pointer to the canonical entry, instead of invisible. Seeded with C10 (whose status here still derives from the new gate: CLOSED with the 9-site baseline recorded). Schema + fixture-repo visibility tests added.

Final receipt: full tree `pytest tests` — **25151 passed, 257 skipped, 6 xfailed, 0 failed**.

silversurfer562:
Record ordering note: registration of C10 is recorded in docs/specs/release-audit-stage/decisions.md via #2269 (chair-read receipt, prior session). This PR is the subsequent Mechanize→Calibrate→Gate step; once merged, the in-repo derived register renders C10 **CLOSED** (gate resolves; the 9-site R8 baseline is recorded) while the canonical register carries pipeline truth: Sweep-fix (the 9 baselined sites) → Close remains. Canonical register doc (~/.attune/reports/attune-ai-review/CLASS-REGISTER.md) updated 2026-08-24: header table row, register-total parenthetical, and pipeline-status line now agree with the derived register.

silversurfer562:
Pushed `ae517d18f` from the C10 sweep-fix session (wonderful-greider): all five Windows lanes were red because `test_rule_flags_an_unregistered_unmarked_read` compares `h.path == "src/mod.py"` while `str(rel)` emits backslashes on Windows. R8+R9 now emit `rel.as_posix()` — hit paths are platform-stable. Verified locally: 238 passed (gates + classes suites). Heads-up to the owning session: this advances the branch head, so re-read before any amend. The sweep-fix PR (baseline → empty) follows once this merges.

PR #2278, opened 2026-08-24
feat(telemetry): pass form_id through the collect handler for stage-latency joins

silversurfer562:
Chair-approved 2026-08-24 (route "1 — measure first"): form telemetry had only `form_surface` + a payload-less `form_submitted`, so per-stage latency was not computable. The implementation lives in attune-forms 0.8.0 (Smart-AI-Memory/attune-forms#59 — `FormSchema.form_id` content-hash join key, `form_build`/`form_rendered`/`form_submitted(form_id)` stage events, `stage_latency()` p50/p95 read-back). This PR is the attune-ai half.

## What's here

- **`_maybe_keyboard_hint(form)`** passes the validated form's `form_id` to `log_submission`, with a `TypeError` fallback so the collect handler works unchanged on attune-forms 0.7.x (zero-arg signature). **No dependency floor bump** — the telemetry lights up automatically when 0.8.0 is installed.
- **Guarded stage tests** through the `attune.telemetry.form_events` alias — `skipif` on the installed capability: skip on 0.7.x (CI today), run on ≥ 0.8.0.
- **Forward-compat fix caught in the act:** three `test_select_form_surface.py` tests counted raw log lines (`len(events) == 1`) and would go red the day 0.8.0 lands (stage events interleave into the same file). They now filter `form_surface` events — the count-assertion drift class from lessons, fixed before it fired.
- **elicit skill Step 0:** the template-store path claim `src/attune/elicitation/templates/` was fiction (that directory has never existed in this repo — templates live in the `attune_forms` package). Fixed, plus the new `template:<name>` vs `dict` cast-source adoption signal named. Mirrors (`.agents/`) and the help reference re-projected.

## Receipts

- Full suite: **25 189 passed, 259 skipped, 6 xfailed** (whole `tests` tree, not per-suite spot runs).
- Dual-mode verification: `tests/unit/telemetry` + `tests/unit/elicitation` + `tests/unit/mcp/test_server_elicitation.py` run **513 passed / 2 skipped** against installed 0.7.0 (CI parity) and **515 passed / 0 skipped** with `PYTHONPATH` pointing at the 0.8.0 source.
- New/changed handler paths covered: normal call shape, TypeError fallback, no-form degradation.

## Template-adoption finding (the chair's investigation ask)

[...]

PR #2280, opened 2026-08-24
fix(capabilities): re-discover plugins in derive_values — kill the polluted-cache flake

silversurfer562:
Second half of today's main-red firefight (#2279 fixed the Windows `SYSTEMROOT` env half; this fixes the flake that remains).

## The flake

`test_real_manifest_repairs_bumped_values_on_copies` fails intermittently across unrelated lanes — ubuntu-3.14 on main's `ea46702f2` run, win-3.12 on docs-only #2277, win-3.10 on #2278 — always the same shape: `mcp_registered_tool_count` derives **50** while the docs say **61** (exactly the plugin tools missing).

## Root cause (verified, not theorized)

`attune.plugins.registry` keeps a module-level `_discovery_cache` and a `_global_registry` singleton. A registry test that runs FIRST on the same xdist worker under mocked `entry_points` leaves an **empty plugin set cached process-wide** (the cleanup conftest only covers `tests/plugins/`). The in-process derivation then builds a server with no plugin tools. The pin: `TestSubprocessCli::test_check_is_green_on_synced_repo` — same env, subprocess, immune by construction — **passed on the very lanes where the in-process test failed**, which rules out the environment and leaves cache state. #2273 adding 243 test lines to `tests/unit/scripts/` reshuffled worker scheduling, which is why it surfaced today.

## Fix

`derive_values()` calls `clear_discovery_cache()` before constructing the server — a derivation describes the checked-out revision, never a prior test's cache. This is squarely inside the module's own charter ("environment-dependent discovery fails closed"; a polluted cache isn't the environment).

## Receipts

- Regression test poisons the cache and asserts derived values match a clean run — **verified red without the fix** (stash/run/pop), green with it.
- `tests/unit/scripts/test_project_capabilities.py`: 48 passed; `tests/unit/gates`: 349 passed.

Unblocks #2277 and #2278 (both currently red on this flake's lanes) once merged into their bases or rerun against green main.

### DIFF
--- a/scripts/project_capabilities.py
+++ b/scripts/project_capabilities.py
@@ -154,6 +154,17 @@ def derive_values(repo: Path) -> dict[str, int]:
                 "tools would be silently missing from the count"
             ) from exc
 
+    # Re-discover from the REAL entry points, always. The registry keeps a
+    # module-level discovery cache and a global-singleton registry; when an
+    # in-process caller (the test suite) runs after a test that touched
+    # either under mocked entry_points, the cached plugin set is wrong and
+    # the derived count silently drops every plugin tool (2026-08-24: 50
+    # vs 61 on whichever xdist worker happened to schedule a registry test
+    # first — the subprocess CLI path never saw it). Derivation must
+    # describe the checked-out revision, not a prior test's cache.
+    from attune.plugins.registry import clear_discovery_cache
+
+    clear_discovery_cache()
     tools = EmpathyMCPServer().tools
     if not isinstance(tools, dict) or not tools:
         raise ProjectionError(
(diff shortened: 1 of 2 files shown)


silversurfer562:
Residual from the parallel chip session's review (verdict: this PR covers the mission at root cause; the chip stood down, no duplicate PR): `TestDerivations::test_registered_count_matches_server_tools` constructs its own `AttuneMCPServer()` in the test process, so the LIVE side of that comparison stays exposed to the same cache pollution if a registry test lands on the worker between the class-scoped `derived` fixture and this test. Two-line hardening: `clear_discovery_cache()` in that test before constructing the server. Inferred from the diff, not reproduced — fold in here or as a follow-up after merge, owner's call.

PR #2300, opened 2026-08-25
feat(chair-arm): block arming a head the local checkout does not hold (chair-read)

silversurfer562:
## Summary

**(chair-read)** — this touches `scripts/chair_arm.py`, which operationalizes D11d CHAIR-ARMS, so it is governance/enforcement surface by construction.

D11d says the chair's label is a read-receipt **bound to a head SHA**. Nothing verified that the head being armed was the commit the operator actually had.

## The incident that motivated it (2026-08-25, same session)

A chair-ruled fix was committed onto a **sibling branch** while I believed I was on the target branch. `git push origin <name>` from a checkout on a different branch **exited 0 having pushed the unchanged ref**, and printed success. #2299 was then armed on a head *without* the fix. It was caught only by hand-comparing `headRefOid` against `git rev-parse HEAD` — nothing in the preflight noticed, and a stale-head arm is indistinguishable from a correct one in every UI.

## The change

`local_head_state(branch, remote_sha) -> (local_sha, relation)` plus a preflight blocker that fires on any relation but `match`:

| relation | meaning | blocker message says |
|---|---|---|
| `ahead` | checkout has commits the PR head lacks — **the incident** | push, then re-run |
| `behind` | stale local ref: the diff you read isn't the diff you'd arm | fetch, then re-run |
| `diverged` | reconcile | reconcile, then re-run |
| `absent` | no local ref — **not a blocker** | (skipped, printed) |

`absent` deliberately fails open: arming from a clone that never fetched the branch stays possible, because there is no local claim to contradict the remote. A head SHA the clone has never fetched is reported `diverged` rather than classified by guess.

## Receipts

- **35 tests pass** (was 30), full tree **25,262 passed**
- Five of the new tests drive `local_head_state` against **real git repositories** (no mocks) covering match / ahead / behind / absent / unknown-SHA — the helper's only job is reading git correctly, so mocking git would test nothing
- **Live dry-run against #2299**: printed `local claude/15-empathy-level-removal matches the PR head` and passed preflight
- **Live incident reproduction**: with a simulated unpushed commit on that branch, the same command BLOCKED —
  ```
  ✗ local claude/15-empathy-level-removal is fd2134fb74ce, PR head is d3d5882112cf
    — your checkout has commit(s) the PR head does NOT — arming here would bind
      the receipt to a head missing that work (push, then re-run)
  ```

### DIFF
--- a/scripts/chair_arm.py
+++ b/scripts/chair_arm.py
@@ -13,9 +13,11 @@ run it only after reading the diff.
 Steps:
  1. Read the PR: state, head SHA, diffstat, changed paths, merge state.
  2. Preflight blockers: open, not draft, base=main, same-repo, not
-    DIRTY (a conflicted PR arms silently and never merges), and no
+    DIRTY (a conflicted PR arms silently and never merges), no
     ``.github/`` paths (the when-green carve-out would disarm + strip
-    the label).
+    the label), and the LOCAL ref for the head branch must match the
+    PR head — arming a head your checkout does not hold binds the
+    receipt to a diff you did not read (2026-08-25 incident).
  3. Name any governance/enforcement surfaces in the diff, so the chair
     knows CHAIR-ARMS class applies (advisory — the read is the control).
  4. Apply the ``auto-merge-when-green`` label (the chair's act).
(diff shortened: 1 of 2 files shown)


silversurfer562:
chair-armed at fe555cde0b5d8170b0a05a6a3e9ab850aa85d697 (merged immediately — already green)

Read-receipt (D11d CHAIR-ARMS): the chair ran scripts/chair_arm.py at head fe555cde0b5d8170b0a05a6a3e9ab850aa85d697 with `auto-merge-when-green` in place. A subsequent push invalidates this receipt — disarm and re-arm after re-reading.

PR #2337, opened 2026-08-27
feat: warn on stale external attune.plugins/attune.wizards entry points (16.0.x)

silversurfer562:
## What

The cheap detector the round table ratified (thread
q-16-release-reliability-001): 16.0.0 collapsed the `attune.plugins`
and `attune.wizards` entry-point groups (release-16-manifest D1, #2334),
so anyone who registered external plugins/wizards through them now fails
by **silent non-loading** — nothing greppable in their own code, so the
migration guide cannot reach them. attune now tells them at startup.

## How

- **`attune.plugins.stale_entry_points`** — one scan of installed
  distribution metadata per process (module-level guard; a failed scan
  is also cached), one warning line per external dist naming the dist
  and the stale group(s), pointing at
  `docs/migration/upgrading-to-16.0.0.md`. Iterates
  `importlib.metadata.distributions()` directly so the owning dist's
  name is available uniformly across Pythons; attune-ai's own dist is
  skipped (its groups are already pinned absent from pyproject by the
  drift guard).
- **Fail-open by contract**: the scan body is wrapped in
  `except Exception` → debug log + return. Broad-except ratchet
  baseline raised `src/attune/plugins/stale_entry_points.py: 1` with
  the reason inline (per the ratchet's escape-hatch protocol).
- **Wired at both once-per-process surfaces**: `PluginRegistry.
  auto_discover()` and the wizard registry's `_load_builtins()` — a
  wizard-only process still gets the warning; the guard prevents
  duplicates.
- **`docs/migration/upgrading-to-16.0.0.md` created** — the guide the
  warning cites did not exist yet (16.0.0 is still Unreleased); covers
  §1 entry-point collapse (the detector's audience) plus the alias and
  root-module removals from the CHANGELOG. No python/json fences with
  removed symbols, so the doc-import audit is unaffected.
- Cleaned three stale "entry-point discovery" comments in
  `wizards/registry.py` left behind by the collapse.

## Receipts

- **Suite**: `pytest tests` (whole tree, xdist) — 24,914 passed,
  259 skipped, 3 xfailed.
- **Unit**: 11 tests in `tests/unit/plugins/test_stale_entry_points.py`
  — per-dist warning, group merge into one line, own-dist +
  nameless-dist skip, quiet-when-clean, fail-open on scan error,
  once-per-process cache (including the failed-scan case), and both
  registry trigger paths. Includes a **non-mocked round trip**: a real
  `*.dist-info` (METADATA + entry_points.txt) on `sys.path`, detected
  through actual `importlib.metadata`.
- **Live-fire**: real process, real dist-info, two
  `PluginRegistry().auto_discover()` calls →

  ```
  WARNING attune.plugins.stale_entry_points: Package 'legacy-ext' registers entry points in 'attune.wizards', which attune-ai 16.0.0 no longer loads — its plugins/wizards are silently ignored. See docs/migration/upgrading-to-16.0.0.md.
  ```

  Exactly one line across both startups.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **Stale entry-point detector**: at the plugin/wizard registries'
+  first load, attune scans installed distribution metadata once per
+  process and logs one warning per external package still declaring
+  entries in the removed `attune.plugins` / `attune.wizards` groups,
+  pointing at `docs/migration/upgrading-to-16.0.0.md`. Without it,
+  such extensions fail by silent non-loading with nothing greppable
+  in the user's own code. The scan is cached and fail-open — a
+  metadata error never affects startup.
+
 ## [16.0.0] - 2026-08-27
 
 16.0.0 is the destructive half of the harness-lite architecture
--- a/docs/migration/upgrading-to-16.0.0.md
+++ b/docs/migration/upgrading-to-16.0.0.md
@@ -48,8 +48,19 @@ The `attune.plugins` and `attune.wizards` entry-point groups are no
 longer read; the dead `attune.workflows` reading path is deleted. The
 bundled plugins and built-in wizards load directly. If you registered
 an external wizard or plugin via these groups, that path is gone —
-the 16.x extension system is the successor. Two things still work
-unchanged:
+the 16.x extension system is the successor.
+
+Because an unread entry point fails by *silent non-loading* (nothing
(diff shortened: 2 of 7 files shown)


silversurfer562:
D11 note: this lead-authored diff touches an enforcement surface (the broad-except ratchet baseline in tests/unit/gates/test_broad_except_ratchet.py, +1 entry with reason inline). Per the collaboration contract that makes it eligible for a different-model review lane before the chair reads the recommendation. Not launched from this session — cross-review is billable and the spend gate needs an explicit go. Say the word and a session can run /cross-review on this PR; the rest of the diff (detector module, tests, migration doc) is outside the risk classes.

PR #2467, opened 2026-09-07
docs(spec): socratic-ambiguity-calibration — per-session asked/inferred/corrected measure; Q4 RULED option 1 (chair, 2026-09-08); D11 lane; G1 status truth-up (chair-read)

silversurfer562:
## What this is

Thread 3 of the 2026-09-07 session ranking ("Socratic calibration metric: amend `docs/specs/socratic-ambiguity-calibration` with an asked / inferred / corrected per-session measure. A candidate, not ruled."). Docs-only; two files in one spec dir plus the R5 ledger row.

Host-surface-parity **D16** (2026-09-07) ruled the dev-session calibration and named this spec as "the place for the product-level version and its measurement". This PR is the measurement half. It was opened as **candidate Q4**; the chair ruled it in chat on 2026-09-08 and the ruling is recorded in the third commit.

## Changes

- **requirements.md** — G5 candidate; new section *Per-session calibration measure*: four counts per session (`asked` / `inferred` / `corrected` / `confirmed_untouched`), two rates read together (correction rate high = under-asking; ceremony rate high = over-asking), a sources table that takes existing stores first per ASI-6 (`form_events.jsonl`, `assumption_review` rulings, `/retro` self-report) and names the two gaps (a PostToolUse hook on `AskUserQuestion`; attune-forms#90 option 3), and boundaries (does not amend the frozen ASI T4 protocol; not a runtime detector; local only; nothing counted is authority).
- **Status truth-up** — G1's rule text shipped 2026-09-07 through D16 + #2459, outside this spec; G2–G4 open. Status stays `parked`; Resume-Trigger is now the Q4 ruling.
- **decisions.md** — Q4 entry: three options, the lead's recommendation (adopt read-only for one week before building anything) and the counter-case carried unprompted.
- **D11 lane, second commit (d812fa4c4)** — Codex seat reviewed 415dc60c1; 2 sent / 0 omitted; 3 findings, all real, all accepted and corrected:
  - F1 the Arm column now reads as a session-level signal; a confirm gate or a D16 trigger-list ask accepted as offered was still owed, and the text says so.
  - F2 one counting unit, the question, across native calls, form fields and typed fallbacks, so neither rate can exceed 1.
  - F3 option 1 now names its coverage limit: native asks have no store and forms have no accepted-untouched flag, so native asks, `confirmed_untouched` and prose inferences are captured by hand at `/retro`, and the week's output is a decision about instrumentation, not a band. The counter-case records that the lane pressed the same point.
[...]

PR #2470, opened 2026-09-08
fix(mcp): version check is single-flight and off under tests — one PyPI thread per process; ATTUNE_VERSION_CHECK=0 opt-out (segfault in CI run 34173042218)

silversurfer562:
## What broke

PR #2467's `coverage` job (run 34173042218, job 101897038684) died with `Fatal Python error: Segmentation fault` in xdist worker gw1 while a docs-only diff was under test. The faulthandler dump shows two threads both inside `ssl.load_default_certs`, both reached from `src/attune/mcp/version_check.py:47 check_for_updates`. Every `AttuneMCPServer(...)` construction started its own daemon thread that opened a live HTTPS request to PyPI; the per-test `server` fixture in `tests/monitoring/test_mcp_path_containment.py` therefore ran several concurrent checks in one worker, and concurrent default-SSL-context creation crashed the interpreter. Coverage itself was 96.45% against the 94% floor. Latent on main for every run; it does not depend on the diff under test.

## Fix

- `version_check.start_background_check()` — lock + started flag, at most one `attune-version-check` daemon thread per process; returns `None` when `ATTUNE_VERSION_CHECK` is `0`/`false`/`no`/`off`. The thread target is resolved at start time, so the existing tests that patch `attune.mcp.version_check.check_for_updates` keep working.
- `server.py` calls it instead of spawning a thread (best-effort guard kept; broad-except ratchet count unchanged).
- `tests/conftest.py` assigns `ATTUNE_VERSION_CHECK=0` at import (every xdist worker inherits it) and suite-manages it so `_scrub_attune_env` leaves it alone. Assigned, not `setdefault` — the Codex lane's finding, below.
- `tests/unit/mcp/test_version_check_single_flight.py` — suite default opts out; four opt-out spellings start nothing; single-flight yields one thread across three starts; two servers share one thread; a server under the suite default never calls PyPI; the patched stub is honored.
- CHANGELOG `Unreleased / Fixed` entry.

Production behavior: the MCP server still checks PyPI once at startup, on one thread. Operators can now switch it off with `ATTUNE_VERSION_CHECK=0`.

## D11 lane (run BEFORE this PR opened; external boundary)

Codex seat (gpt-6-astra per `~/.codex/config.toml`; ChatGPT subscription login, no API spend), staged diff, 5 files sent / 0 omitted, Board thread `review-fix-mcp-version-check-single-flight-20260908-0042`. **1 finding, real, accepted and corrected:** `os.environ.setdefault` would have preserved an inherited `ATTUNE_VERSION_CHECK=1` from a developer shell, and the suite-managed exclusion would then have protected it from the scrubber. Now assigned unconditionally; proved by running the focused suites under `ATTUNE_VERSION_CHECK=1` (56 passed). Ledger row appended to `docs/specs/cross-review/receipts.md`.

## Receipts (run locally on this head)

| Probe | Result |
| --- | --- |
| new tests + `test_version_check*` + MCP memory/help handlers + path containment + workspace + run-record corpus | 134 passed |
| `tests/unit/gates` + `tests/unit/ci` (broad-except, sys.modules, workflow ratchets) | 1,041 passed, 1 existing skip |
| `version_check.py` statement + branch coverage (/tmp recipe) | 100% |
| focused suites under an inherited `ATTUNE_VERSION_CHECK=1` | 56 passed |
| whole configured tree, keyless (`ANTHROPIC_API_KEY=""`), worktree venv 3.11 | 26,097 passed, 266 skipped, 3 xfailed in 108.73s; 0 failed |
| pinned pre-commit on all changed files, R5 ledger gate included | passed |

## Disclosures

- The whole-tree run started before the lane's one-line conftest change (`setdefault` → assignment); with the variable unset in that shell the two lines behave identically, so the receipt covers the same code path. Stated, not re-run.
- Windows lanes are not certified locally; the change is pure Python threading and env, no path or subprocess handling.
- Not addressed here: whether the MCP server should phone PyPI at startup at all (the telemetry-local-only stance). Kept as-is; the opt-out is new.


silversurfer562:
Rebased onto main after #2467 merged (both PRs appended an R5 ledger row; both kept, content otherwise unchanged, commit re-verified signed). New head d885a1289, confirmed on the remote; CI re-running. Merge that head when green; nothing armed.

PR #2474, opened 2026-09-08
fix(release-tooling): three chair-ruled retro fixes (chair-read)

silversurfer562:
## Summary

The three chair-ruled do-now fixes from the 2026-09-08 retro, held for after the 16.3.0 release per the chair's "do them a separate branch to be done after the release."

**1. `scripts/consolidate_changelog.py`** — PRs append their own `###` header under `[Unreleased]` instead of inserting under the existing one, so cutting 16.3.0 found seven headers in one section (2 Added, 2 Changed, 3 Fixed) and they were merged by hand. Now an order-preserving, idempotent transform with a `--check` mode and a drift guard that fails at PR time rather than release time.

**2. `review.ledger_row()` self-classifies a clean lane.** `not-triaged` is not a legal disposition — `ledger_precision`'s tally rejects any row it cannot classify — so the module's own output failed the repo's own pre-commit gate, and every clean lane cost a hand-edit. A clean lane over a PARTIAL manifest now says the omission in the row itself. `absent` and `format_noncompliant` lanes keep the placeholder: they also carry zero findings, but they judged nothing, and auto-classifying one as clean would launder a non-review into a receipt.

**3. `reach_snapshot.py --out` defaults to the MAIN checkout** (via `git rev-parse --git-common-dir`). Run from a worktree the old relative path landed the file inside that worktree, invisible to the main checkout's `git status` and lost when the worktree is reaped — which is how the 16.3.0 snapshot nearly went missing. Degrades to the relative path when git is unavailable, errors, or the directory does not exist; a scheduled job must not crash on this.

## Two defects found while building, not after

- **Own test caught one**: when the target section was last in the file, the rebuild appended a stray blank line, so a clean changelog would be mutated and `--check` could never report success. Fixed before commit.
- **The cross-review lane caught the other** (medium, confirmed by probe before accepting): heading detection ignored code fences, so an unindented `### Added` inside a fenced example was read as a category boundary and **silently deleted from inside its fence** — data loss in a script that writes `CHANGELOG.md`. A changelog entry naturally quotes changelog syntax, so it is reachable. A fence mask now gates every heading test in both the section scan and the category split; unbalanced fences mask the remainder and fail safe. Four fixtures pinned: the reported case, a fenced `## [x]` release heading (which truncates the section instead), an indented fence, an unbalanced fence.

## Verification

| Claim | Probe | Result |
| --- | --- | --- |
| Whole tree, keyless, final head | `ANTHROPIC_API_KEY="" pytest tests -n auto` | 26,143 passed, 266 skipped, 3 xfailed, 0 failed |
| Changed module coverage | `coverage --source=attune.roundtable.review` | 99% |
| Consolidator behavior | `tests/unit/scripts/test_consolidate_changelog.py` | 20 tests |
| Ledger rows round-trip the gate | clean / clean-on-partial / absent / format_noncompliant through `check_disposition` | all pass; absent keeps the placeholder |
| Snapshot path from a worktree | real `git worktree add` fixture | worktree and main resolve to the same directory |
| Ledger gates after the R5 row | `test_ledger_precision.py` + `test_ledger_rejection_format.py` + the pre-commit hook | 26 passed, hook passed |
| Pinned hooks | `pre-commit run --files <all changed>` | all passed |
| Commits signed | `git log --format=%G?` | G on all three |

## D11 lane

Codex seat, run BEFORE this PR opened (release tooling and the R5 ledger emitter itself). Complete manifest: 6 sent, 0 omitted, 22,696 chars. One real finding, accepted and fixed in-branch; R5 row appended. Subscription login, no API spend.

**Disclosure**: two commits land after the lane read the diff — `59d2e40cc` (the fence fix the lane asked for, plus its four fixtures) and `d12003929` (the ledger row itself). The whole tree was re-run on the final head after both.

## Note

`plugin/skills/cross-review/SKILL.md` still lists `not-triaged` among the dispositions offered to the user. That stays accurate for lanes with findings and is left alone to keep this diff scoped; a skill edit would also require re-running the agents-skills projection.


silversurfer562:
## Changelog gate satisfied (head `78ee1a817`)

The required `changelog-entry` check failed on the first push — correctly: this ships a `src/` change and carried no entry. Added one under `[Unreleased]` rather than applying `no-changelog`, since anyone running `/cross-review` sees a different (and now legal) ledger row.

Dogfooding the new script on that entry surfaced a message bug in it: `--check` printed "has duplicate ### headers" when the section's headers were each unique and only their **order** was non-canonical. It now names both possible reasons, and the section was normalized by running the script itself.

Post-lane commits now stand at four (D13): `59d2e40cc` fence fix, `d12003929` ledger row, `78ee1a817` changelog entry + message fix. All docs/tooling; the only `src/` change remains the `ledger_row()` fix the lane already read.

PR #2512, opened 2026-09-11
fix(wizards): parse task blocks one at a time when the whole span is not well-formed (stacked on #2511)

silversurfer562:
## What

Stacked on #2511 (base `fix/read-spec-elementtree`, one commit on top of its head `3a34a29db`). When the task-XML span is not well-formed as a whole, `TaskDecomposer._parse_tasks_from_xml` now retries each `<task …>…</task>` block on its own with defusedxml and hands only the blocks the parser rejects by themselves to the regex extractor. The whole-span parse stays first; the regex path stays the final fallback, unchanged, with every one of its warnings. This is the follow-up chipped from the third codex lane on #2511 (rejected there as a defect, accepted as a recall improvement) — opened as information for the #2511 review, not as a change to that diff.

## Why

One bare `&` in prose between blocks, or one broken task among many, sent every task to the regex path, where single-quoted, reordered or extra attributes drop a task (the orphan scan names it, but the task is still gone). Per-block retry keeps the parser for every block that is well-formed on its own.

## The two numbers for the chair

**(a) Corpus, every tracked `*.md` containing `<task id=` (49 files), `to_dict()` field for field:**

| | #2511 head | this PR |
|---|---|---|
| files on the parser path as a whole | 34 | 34 |
| files that fall back | 15 (whole file to regex) | 15 (per block: parser for the well-formed blocks, regex for the rest) |
| of those 15, files with every block recovered by the parser | — | 0 (in all 15 the malformed fragment sits inside a block, not between blocks) |
| identical / differing vs the regex-only path | 42 / 7 | 37 / 12 |
| identical / differing vs the #2511 head | — | 44 / 5 |

The 5 new differences (the 12 minus #2511's 7), classified programmatically — 9 entity decodes, 1 recovered file entry, 0 other:

- `.claude/plans/feature-rag-code-grounding-2026-04-17.md`: task 3.2 has two `<files-to-create>` sections; the regex read only the first and warned "4 present / 3 parsed". The parser reads both: `packages/attune-author/tests/unit/test_rag_hook.py` recovered, warning resolved.
- `.claude/plans/help-docs-rollout-gate.md` (T1.objective, T1/T2.files_to_modify), `.claude/plans/spec-gate-real-review.md` (task 1 checks), `docs/specs/archive/bulletin-curator/tasks.md` (task 3.3 files), `docs/specs/host-surface-parity/tasks.md` (tasks 3, 5, 6): `&lt;`/`&gt;` written as escapes now decode in the blocks that are well-formed on their own — the divergence #2511 already pinned.
- Warnings across the corpus: 0 lost, 0 gained, after classifying the two that disappeared — the feature-rag one was resolved by the recovery above; `docs/specs/test-discipline-controls/tasks.md` task 1.1's "4 present / 2 parsed" counted two `<file>` tags in a non-schema `<files-to-read>` section, a false positive the parser's exact section count removes (same as #2511 for any parser-parsed task).

**(b) Net lines on top of #2511:** `src/attune/wizards/decomposer.py` **+38 / −11** (#2511 itself is +135 / −12 on main); tests +66 / −2; CHANGELOG +6.

## Other receipts (this machine, worktree `inspiring-volhard-365a2b`, head a787b5480)

- `tests/unit/wizards/test_wizard_decomposer.py` + `tests/unit/pipeline/test_spec_reader.py`: **80 passed** serial (`-p no:xdist`); `tests/unit/wizards` + spec_reader: **486 passed** serial. 5 new tests: bare `&` between blocks keeps the parser for a single-quoted, reordered, `depends-on=` task; only the rejecting block falls back, document order kept across the mix; an unclosed task still yields the regex swallow warning per block while the task after it keeps the parser; orphaned content between blocks is still named with the parsed count; a block that fails both parsers is named and counted as dropped.
- `bandit -c .bandit --severity-level medium --confidence-level medium` on the module: clean. Pinned black + ruff `--no-fix`: clean; full pre-commit on the three files: clean. `tests/unit/gates`: 671 passed.
- Whole tree (`pytest tests`, xdist): **26,612 passed**, 323 skipped, 6 xfailed, 0 failed, 6 errors — all six the setup fixture of `tests/memory/test_ams_backend_integration.py`, which probes a local Agent Memory Server that is not running on this machine (environmental, unrelated to this diff) on a787b5480.

## Merge note

If #2511 squash-merges first, GitHub closes this stacked PR; the single commit then cherry-picks onto a fresh branch off `main` and reopens (its CHANGELOG hunk anchors on #2511's entry). If the chair prefers the change folded into #2511, say so — nothing is pushed to `fix/read-spec-elementtree` by default.

## D11 disclosure

Core spec-execution parsing — not a listed risk class (same call as #2511). No cross-model lane run on this diff.


silversurfer562:
Chair ruling (2026-09-11): **closed, not shipped.** The receipts were re-run centrally and matched exactly (44/49 identical to #2511, 5 in the parser's favor, +38/−11 src). Ruled against because per-block parsing moves zero files off the regex fallback, pushes entity decoding to block granularity inside a single plan, and recovers one section in one plan for its size; the two-path split is inherent to any fallback design. The parser it stacked on landed in #2511 (`6c881483b`). Not to be re-opened without a new reason.

silversurfer562:
Re-ruled with the author session's counter-case in front of the chair (the corpus is hand-written plans; the target is LLM decomposition output where a bad fragment and a quoted/reordered task can co-occur): **still closed, but the branch is parked** — `fix/decomposer-per-block-parse` stays on origin with `a787b5480` as insurance. Reopen only with evidence from a real decomposition run, not the hand-written corpus.

PR #2516, opened 2026-09-12
fix(doctor): optional extra raising a non-ImportError no longer crashes attune doctor; scope the metadata patch in its tests (salvaged from session 3)

silversurfer562:
## What this is

Salvage of real, uncommitted work from CCD session "still open session 3" (worktree `objective-mahavira-cad2f1`, 2026-09-09), applied verbatim on a fresh branch off `origin/main`. That worktree's three dirty files are **left untouched** as the backup until this merges; nothing there was committed, reset, or reaped.

## Two separable defects, both fixed

**Product.** `attune doctor`'s optional-extra probe caught only `ImportError`. An extra whose import raises anything else at import time (a missing native library, a CUDA probe, an import-time `importlib.metadata` read) took the whole diagnostic down with a traceback. It now reports `[--] <pkg> import failed: <e>`, kept distinct from "not installed". The broad-except ratchet baseline moves 7 → 8 with the reason recorded in the test.

**Test.** `test_related_packages_broken_metadata` blanket-patched `importlib.metadata.version` for every package. `huggingface_hub` calls that function at import time, so on any interpreter with `sentence-transformers` installed the mock fired inside third-party code and the test crashed. CI never installs that extra (it is absent from `pyproject.toml` and `uv.lock`), so CI was green by accident of a leaner environment. The patch is now scoped to the three `attune-*` names and delegates everything else to the real function; `sentence_transformers` is pinned absent in the fixture; `test_optional_extra_import_raises_non_import_error` pins the new path.

## Receipts

| Probe | Result | Where |
| --- | --- | --- |
| Reproduction on pyenv 3.10.11 (has the extra) | identical traceback, `Error importing huggingface_hub.hf_api: boom` | session 3, relayed |
| Whole tree, pyenv 3.10.11 / `.venv` 3.11.14 | 26,163 passed / 26,144 passed, raw exit 0 | session 3, relayed |
| `tests/unit/cli_commands/test_utility_commands.py` + `tests/unit/gates/test_broad_except_ratchet.py` against THIS worktree's source (`PYTHONPATH` override, not main's editable mapping) | 51 passed | this head, re-run |
| Mutation: product fix reverted → new regression test | 1 failed (guard is real) | this head, re-run |
| Diff identity vs session 3's worktree (index lines aside) | identical | this head |
| Pinned pre-commit hooks on the three files | passed | this head |
| `tests/unit/gates` full run (worktree source) | 672 passed | this head, re-run |

## Disclosures

- The ratchet baseline is an enforcement surface; the bump is +1 with its reason in the same commit. Whether that needs a D11 lane or merges on receipts is the chair's call (09-06 waiver).
- Session 3 also found that `pytest … | tail` reports `tail`'s exit status, not pytest's; that and the blanket-patch class are one lesson now in the docs outbox (`20260911-2105-lesson-blanket-stdlib-patch-bleeds-into-third-party-imports.md`), not appended to `.claude/lessons.md` here.
- Authored by session 3 (Claude); landed by the next-session-start session; the salvage was claimed with "still open session 1", which had the same brief, before starting.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Plugin metadata lookup now handles registered engine workflows without
   constructing them; bundled software workflow lookup no longer crashes on
   legacy analyzer-only fields.
+- `attune doctor` no longer crashes when an optional extra raises
+  something other than `ImportError` at import time (a missing native
+  library, a CUDA probe, an import-time metadata read); the package is
+  reported as `import failed`, distinct from `not installed`.
 
 - Class scans and class-register commands now return a non-zero exit code
   when a Python file cannot be parsed. The parse diagnostic remains in the
--- a/src/attune/cli_commands/utility_commands.py
+++ b/src/attune/cli_commands/utility_commands.py
@@ -471,6 +471,14 @@ def cmd_doctor(args: Namespace) -> int:
             _ok(f"{pkg_name} installed")
         except ImportError:
             _warn(f"{pkg_name} not installed", "optional")
+        except Exception as e:  # noqa: BLE001
+            # INTENTIONAL: a diagnostic must never crash on a broken extra.
+            # Import-time failures that are NOT ImportError are routine for
+            # this class of package — a missing native library, a metadata
+            # read, a torch/CUDA probe — and "installed but unimportable" is
+            # a different fact from "absent". Report which one, never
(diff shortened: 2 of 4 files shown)


silversurfer562:
Changelog entry added under Unreleased → Fixed (the changelog-entry gate was the one failing check). The four recovery PRs each insert after a different existing bullet so they merge cleanly in any order; if a rebase is still needed after a sibling lands, it is a keep-both resolution. Head now 89b238cd1.

PR #2517, opened 2026-09-12
fix(hooks): guard tokenizer splits on newlines and treats heredoc bodies as data; worktree-add guard matches the subcommand after git's global options (stranded-diff recovery 1/3)

silversurfer562:
## What this is

PR 1 of the stranded-diff recovery (Patrick, 2026-09-11: "recover the four stranded diffs as four PRs, one at a time on fresh branches from main in this worktree, tests run per PR, awesome-noether first; the owning worktrees stay untouched"). The doctor fix was #2516; this is the guards fix from worktree `awesome-noether-f81a8d` (CCD session "Make worktree_add_guard match the git -C form", tonight), applied verbatim on a fresh branch off `origin/main`. That worktree's four dirty files are **untouched** and remain the backup until this merges.

## The defect

The 2026-09-11 escape was a multi-line Bash script: `set -e`, two assignments, then `git -C <main> worktree add …` on its own line. The shared tokenizer split invocations only on `&&` `||` `;` `|` `&`, never on newlines, so the whole script became one invocation starting with `set`, which the git detector rejected. The single-line `-C` form was already refused and already pinned. Verified by the authoring session by replaying the exact command from the good-morning-247b52 transcript through the unmodified script from a worktree cwd: exit 0.

## The fix

- **`dirty_switch_guard.py`** (owner of the shared tokenizer): newline is a command separator; backslash-newline joins; `#` comment handling is disabled so a trailing comment cannot swallow the next line; here-document bodies are stripped first so a heredoc containing `git worktree add` or `git checkout main` is data, not a command. Without the heredoc step, the newline split would have made the dirty-switch guard block ordinary heredoc file writes from a dirty tree. Reaches all three consumers: worktree-add, dirty-switch, detached-head-push.
- **`worktree_add_guard.py`**: `is_worktree_add` skips git's global options (`-C`, `-c`, `--git-dir`, `--work-tree`, `--namespace`, `--config-env` with values; other leading flags without) and matches `worktree add` by position, so `git commit -m worktree add` is not a worktree add. Escape hatch and main-checkout allowance untouched.
- **Tests** pin the bare, `-C`, `-c … -C`, `--git-dir=… --no-pager`, `--work-tree`, exact 09-11 multi-line, continuation and trailing-comment forms as refused; `list`, `remove`, `-C … list`, `commit -m worktree add` and a heredoc body as allowed; the escape hatch over both forms; a next-line switch pin and a heredoc-is-data pin on the dirty guard. No new `except`; the broad-except baseline is unchanged.

## Receipts on this head (worktree source via `PYTHONPATH`, not main's editable mapping)

| Probe | Result |
| --- | --- |
| Both guard test modules, serial | 93 passed |
| Whole `tests/unit/hooks`, serial | 1152 passed |
| `tests/unit/gates` + `tests/unit/plugins/test_sdk_subprocess_gate.py` (every repo hook gated) | 715 passed |
| Live stdin, `HOME` redirected to a scratch dir: exact 09-11 multi-line, single-line `-C` | exit 2 (refused) |
| Live stdin: heredoc body only (both guards), `git worktree list`, `git commit -m worktree add` | exit 0 (allowed) |
| Diff identity vs the source worktree (index lines aside) | identical |
| `git apply --check` of the source diff against `origin/main` via a temp index | clean |
| Pinned pre-commit hooks on the four files | passed |

The authoring session's own receipts (93 / 1152 / 252, ruff on two versions, black, live probes) are in its transcript and were not relayed as this PR's evidence; everything above was re-run here.

## Disclosures

- **Enforcement surface.** These are the two PreToolUse guards that run on every Bash call in a session; a wrong fix blocks or unblocks git fleet-wide. Whether that warrants a D11 lane or merges on receipts is the chair's call.
- **Metrics log.** The authoring session's live probes appended ~7 "fired" and 2 "allowed" rows to `~/.attune/enforcement-metrics.jsonl` (2026-09-12T00:0x–00:14Z, detail = its worktree path) and was blocked from removing them; they will inflate the worktree-add-guard hit count at the next retirement review unless Patrick strips them. This PR's probes wrote to a scratch `HOME` instead.
- Landed by the next-session-start session; claimed with "still open session 1" (which found the diff) and announced to the authoring session before push.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   when a Python file cannot be parsed. The parse diagnostic remains in the
   output, and scanning continues for other files; an incomplete scan can
   no longer silently pass the register's scan-error check.
+- The worktree-add, dirty-switch and detached-head-push guards split
+  multi-line Bash commands on newlines, treat here-document bodies as
+  data, and match `git worktree add` by position after git's global
+  options (`-C`, `-c`, `--git-dir` …): a `git -C <main> worktree add`
+  on the line after `set -e` is now refused, while `git commit -m
+  worktree add` and heredoc bodies are not.
 - `/cross-review` no longer briefs the authoring seat on its own diff
   when the moderator is not Claude. `run_review()` resolves an omitted
   `seat` against the moderating host: Claude-hosted and host-less runs
--- a/src/attune/hooks/scripts/dirty_switch_guard.py
+++ b/src/attune/hooks/scripts/dirty_switch_guard.py
@@ -53,6 +53,7 @@ from __future__ import annotations
 
 import json
 import os
+import re
 import shlex
 import subprocess
 import sys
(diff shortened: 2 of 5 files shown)


silversurfer562:
Changelog entry added under Unreleased → Fixed (the changelog-entry gate was the one failing check). The four recovery PRs each insert after a different existing bullet so they merge cleanly in any order; if a rebase is still needed after a sibling lands, it is a keep-both resolution. Head now 0974a35d0.

PR #2518, opened 2026-09-12
fix(ops): session_summary_cache validates session_id against the flat cache dir; drop its allowlist entry (stranded-diff recovery 2/3)

silversurfer562:
## What this is

PR 2 of the stranded-diff recovery (Patrick, 2026-09-11: "recover the four stranded diffs as four PRs, one at a time on fresh branches from main in this worktree, tests run per PR; the owning worktrees stay untouched"). #2516 was the doctor fix, #2517 the guards; this is the session-summary-cache traversal guard from worktree `morning-cleanup-review-1150e7` (CCD session "Validate session_id in ops session-summary cache path", 2026-08-21 — the chip filed by the 08-21 path-validation lesson), applied verbatim on a fresh branch off `origin/main`. That worktree's two dirty files are **untouched** and remain the backup until this merges.

## Reachability (the authoring session's trace, re-stated)

`cache_path_for()` is called only by `load()`/`save()` in the same module; those are called only from `session_summarizer.py`, where `session_id = jsonl_path.stem` of a locally globbed `*.jsonl`; `summarize_session()` has zero non-test callers in `src/` (the dashboard routes that once forwarded ids were deleted in #1545). A stem cannot carry a separator, so **no live traversal existed**. The guard is defence in depth for future callers of a public helper whose docstring already promised a flat `<id>.json` layout.

## The change

- `cache_path_for()` checks the flat-directory invariant first (the composed path's parent must be the cache dir, rejecting escaping **and** merely nesting ids), then routes through `_validate_file_path` for null bytes and resolution-based containment. The validator's resolved value is discarded so a symlinked `attune_home` is not rewritten under callers. Creates nothing.
- `save()` raises before `mkdir`, `mkstemp` and the rename, so a bad id touches nothing; `load()` raises rather than treating a traversal as a cache miss. Docstrings carry the `ValueError` contract.
- 9 regression tests: escaping ids, an absolute id, a nesting id (`sub/dir/x`), a no-filesystem-mutation assertion on `save()`, and pins that real UUID ids and the `..` stem still work.
- **Allowlist ratchet:** the module now references the validation helper, so `test_allowlist_entries_are_still_needed` failed on the applied diff (1 failed / 10 passed) until its entry and the NOTE that tracked exactly this question were removed from `tests/unit/gates/test_path_validation_gate.py`. The 08-21 re-seed group header stays because the other re-seeded modules remain listed. (The entry did not exist when the session wrote the diff; it was re-seeded later that day.)

## Receipts on this head (worktree source via `PYTHONPATH`)

| Probe | Result |
| --- | --- |
| `tests/unit/ops/test_session_summary_cache.py` | 26 passed |
| Whole `tests/unit/ops` | 1688 passed |
| `tests/unit/gates` (path-validation gate included, entry dropped) | 672 passed |
| Mutation: source guard reverted → new tests | 10 failed / 16 passed (the guard is real) |
| Diff identity vs the source worktree (index lines aside) | identical |
| `git apply --check` against `origin/main` via a temp index | clean |
| Pinned pre-commit hooks on the three files | passed |

The authoring session's own live-fire (a real `save()` with `../../../escaped` raising before any write under a temp home) is in its transcript and not relayed as evidence here.

## Disclosures

- Security-adjacent path handling plus a gate allowlist change: D11 risk classes. Lane or merge-on-receipts is the chair's call.
- The authoring session flagged a separate gap: about 12 modules use `mkstemp`/`NamedTemporaryFile` and were invisible to the scanner at the time. Main's scanner has since learned that idiom (the 08-21 re-seed), so that gap is closed; nothing rides along here.
- Landed by the next-session-start session; claimed with "still open session 1"; the authoring session is idle since 08-21 and was not messaged (absence, per the peer-driving protocol) — its worktree was read, never written.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   including the moderator's own; that run is stamped `self_review: True`
   rather than passed off as a cross-review. Saved board posts and ledger
   rows retain the host, self-review state, and resolved Claude auth route.
+- The ops session-summary cache validates its session id against the
+  flat cache directory before any write: an id that would escape or
+  nest raises `ValueError` and touches nothing. No shipping caller could
+  reach a traversal; this is defence in depth.
 - `claude_auth` defaults to `"auto"`, which selects the verified Pro/Max
   subscription route ONLY for a real `claude` seat briefed from a known
   non-Claude host. That cross-host case previously failed with
--- a/src/attune/ops/session_summary_cache.py
+++ b/src/attune/ops/session_summary_cache.py
@@ -29,6 +29,8 @@ from dataclasses import asdict, dataclass
 from datetime import datetime, timezone
 from pathlib import Path
 
+from attune.security.path_validation import _validate_file_path
+
 logger = logging.getLogger(__name__)
 
 # Bytes hashed from the JSONL tail to form the cache-key digest.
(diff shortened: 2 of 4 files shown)


silversurfer562:
Changelog entry added under Unreleased → Fixed (the changelog-entry gate was the one failing check). The four recovery PRs each insert after a different existing bullet so they merge cleanly in any order; if a rebase is still needed after a sibling lands, it is a keep-both resolution. Head now 171595c2b.

PR #2519, opened 2026-09-12
fix(llm): thread max_tokens through EmpathyLLM.interact and the level handlers; executor mixin and test-gen pass their caps (stranded-diff recovery 3/3)

silversurfer562:
## What this is

PR 3 of 3 of the stranded-diff recovery (Patrick, 2026-09-11: "recover the four stranded diffs as four PRs, one at a time on fresh branches from main in this worktree, tests run per PR; the owning worktrees stay untouched"). #2516 doctor, #2517 guards, #2518 traversal guard; this is the `max_tokens` threading from worktree `context-management-792e27` on `fix/llm-max-tokens-threading` (2026-08-19; its session is no longer listed), applied on a fresh branch off `origin/main`. That worktree's four dirty files and one untracked test are **untouched** and remain the backup until this merges.

## The defect

`EmpathyLLM.interact()` had no `max_tokens` parameter. Every level handler used `EmpathyLevel.get_max_tokens_recommendation(<level>)` unconditionally (1024–4096), so a caller with a long output, such as test or doc generation, was silently truncated at the level default with no way to raise it.

## The change

- `EmpathyLLM.interact(max_tokens: int | None = None)`: documented override, passed to `_level_1_reactive` … `_level_5_systems`; each handler takes `max_tokens` and uses it ahead of the level recommendation.
- `workflows/executor_mixin.py`: a step's `max_tokens` is forwarded into the call kwargs via `setdefault`, so an explicit kwarg still wins.
- `workflows/test_gen_parallel.py`: passes `max_tokens=16384` for its generation step.
- `tests/unit/workflows/test_llm_max_tokens_threading.py` (new): pins the threading at each seam.

## Fidelity to the source diff

- One hunk offset in `executor_mixin.py` (`@@ -216` vs `@@ -186`): the later #2314 refactor added 30 lines above the touched region. Content identical; `git apply --check` against `origin/main` via a temp index was clean.
- **Black (pinned) rewrapped seven long lines**: the five `_level_N(...)` calls in `llm/interaction.py` and two `patch.object(...)` lines in the new test. Formatting only; no token changed. The source worktree's copy predates the current pin.

## Receipts on this head (worktree source via `PYTHONPATH`)

| Probe | Result |
| --- | --- |
| New test + `tests/unit/llm` + `test_executor_mixin_hybrid.py` + `test_parallel_test_gen.py`, serial | 102 passed |
| `tests/unit/workflows` + `tests/unit/models`, xdist | 3708 passed, 1 skipped |
| Mutation: threading reverted in `llm/core.py` + `llm/interaction.py` → new test | 1 failed / 3 passed (the pin is real) |
| New `except` clauses | 0 |
| Pinned pre-commit hooks on the five files (after black's rewrap) | passed |

## Disclosures

- Touches the LLM call path (`attune.llm`), a models/workflows seam that #2314 was cleaning up; no import direction changes here, but a reviewer of that layering may want a look. Lane or merge-on-receipts is the chair's call.
- No API-key call was made; every run used `ANTHROPIC_API_KEY=""`.
- Landed by the next-session-start session; claimed with "still open session 1"; the authoring session is absent (not in the session list), so no owner was messaged; its worktree was read, never written.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   raises `ReviewTargetError`. An unknown host still resolves to `"api"`,
   preserving the zero-cap refusal. Pass `claude_auth="api"` to force the
   billable route.
+- `EmpathyLLM.interact()` accepts `max_tokens`, threaded through all
+  five empathy levels; the workflow executor forwards a step's cap and
+  parallel test generation passes 16384, so long generated outputs are
+  no longer silently truncated at the level default.
 - The CI Ruff gate now runs `ruff check --no-fix .`. With `fix = true` in
   `pyproject.toml`, the bare `ruff check .` rewrote the runner's checkout and
   exited 0, so every auto-fixable violation passed the gate and merged
--- a/src/attune/llm/core.py
+++ b/src/attune/llm/core.py
@@ -310,6 +310,7 @@ class EmpathyLLM(SecurityMixin, InteractionMixin):
         context: dict[str, Any] | None = None,
         force_level: int | None = None,
         task_type: str | None = None,
+        max_tokens: int | None = None,
     ) -> dict[str, Any]:
         """Main interaction method.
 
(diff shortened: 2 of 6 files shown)


silversurfer562:
Changelog entry added under Unreleased → Fixed (the changelog-entry gate was the one failing check). The four recovery PRs each insert after a different existing bullet so they merge cleanly in any order; if a rebase is still needed after a sibling lands, it is a keep-both resolution. Head now 324fb1ba0.

PR #2522, opened 2026-09-12
feat(scripts): salvage_worktree_diff.sh — read-only salvage of a stranded worktree diff onto a fresh branch (retro 2026-09-11 item 1)

silversurfer562:
## What this is

Retro 2026-09-11 item 1 (chair-ruled do-now). Mechanizes the
stranded-diff salvage recipe that ran by hand 4/4 tonight
(#2516–#2519): lift a worktree's uncommitted diff verbatim onto a
fresh branch cut from current main, in the current checkout, and
leave the source worktree untouched as the backup until the PR merges.

```text
scripts/salvage_worktree_diff.sh <source-worktree> <new-branch> \
    [--base <ref>]            # default base: origin/main
```

Exit status: `0` branch created + patch applied + identity held;
`1` refused (inside the source, dirty checkout, fetch/checkout failed,
patch does not apply, untracked file collides with a file here,
identity differs); `2` nothing to salvage; `64` usage.

## Behavior, in order

1. Refuses to run inside the source worktree (toplevel comparison)
   or from a dirty checkout (`git status --porcelain` non-empty).
2. Read-only capture: `git diff --binary HEAD` (staged AND unstaged)
   into a temp patch, plus the `??` lines of
   `git status --porcelain -z --untracked-files=all`. Both empty
   -> exit 2. The source reads run under `GIT_OPTIONAL_LOCKS=0`, so
   git does not even refresh the source's index.
3. `git fetch <remote> <branch>` derived from `--base`, then
   `git checkout -b <new-branch> <base>` — after `cd` to the
   toplevel, because `git apply` from a subdirectory silently
   ignores paths outside it.
4. `git apply --check`, then apply; each untracked file is copied to
   the same relative path (dirs created, `cp -p`), refusing with the
   path if anything already exists there.
5. Identity receipt: `diff` of the two patches with `index` lines and
   `@@` hunk headers stripped — prints `IDENTICAL` or the differing
   lines (exit 1). Then the receipt: `--stat` (files, +/- counts),
   untracked copies, the untouched-backup reminder, next steps
   (tests against THIS worktree's src, commit via `-F`, push, PR).

bash-3.2 safe (no `mapfile`, empty arrays guarded) so `/bin/bash` on a
Mac without Homebrew bash runs it; shellcheck clean.

## Tests

`tests/unit/scripts/test_salvage_worktree_diff.py` builds a throwaway
repo (primary checkout + bare `origin` with `main` pushed + a linked
worktree), dirties the worktree with a tracked edit and a nested
untracked binary file, and drives the real script via `bash` from the
primary checkout. Cases: happy path (branch created, normalized diff
identical, untracked copied byte-for-byte, source status/HEAD/every
file byte unchanged, exit 0); inside the source -> 1; dirty current
checkout -> 1; clean source -> 2; five usage shapes -> 64; untracked
file colliding with a tracked file on the base -> 1 with the path and
nothing applied. Fixture git calls carry `-c commit.gpgsign=false`
under a scratch `GIT_CONFIG_GLOBAL`/`GIT_CONFIG_SYSTEM`/`HOME`.

## Receipts (what was actually run, this head)

| Probe | Result |
| --- | --- |
| `uvx --from shellcheck-py shellcheck scripts/salvage_worktree_diff.sh` | rc=0, no findings (ShellCheck 0.11.0) |
| `tests/unit/scripts/test_salvage_worktree_diff.py`, serial (`-p no:xdist -o addopts=`), `PYTHONPATH=<this worktree>/src`, main venv py3.10, `ANTHROPIC_API_KEY=""` | `10 passed in 2.67s` |
| Same file with `PATH=/usr/bin:/bin:/usr/sbin:/sbin` so `bash` is `/bin/bash` 3.2 | `10 passed in 3.75s` |
| `tests/unit/scripts`, serial, same env | `601 passed in 22.50s` |
| `uv run --frozen --with pre-commit pre-commit run --files <both>` | every hook Passed (black, ruff, bare-except, bandit, detect-secrets, whitespace, eof, line-ending, brand-drift); no reformat |
| `git log -1 --format='%G?'` / `git status --short` after commit | `G` / empty |

Basis note: the pytest counts are read from pytest's own summary line;
those runs were piped through `tail`, so the exit status observed was
`tail`'s, not pytest's.

## Disclosures

- **No lane owed.** `scripts/` is not an enforcement surface (not a
  gate, guard, ledger, or contract text); no `src/` or gate changes.
- **Recipe source:** the four recovery PRs tonight (#2516–#2519) —
  applied verbatim on a fresh branch off `origin/main`, identity
  "index lines aside", source worktree left untouched as the backup.
- **Deviation from the brief's literal `git diff`:** the capture is
  `git diff --binary HEAD` (staged AND unstaged). A bare `git diff`
  drops anything the stranded session had `git add`ed, and the salvage
  would still print IDENTICAL — a receipt that lies is the worst
  failure mode here. The identity check compares `git diff --binary
  HEAD` on both sides; with nothing staged on the new branch that is
  the brief's `git diff`.
- **Collision rule is stricter than "tracked":** the copy refuses if
  ANY file exists at the destination. The checkout was verified clean,
  so anything there is tracked or ignored; neither is clobbered.
- **Paths not exercised by tests:** `--base` with a non-remote ref,
  the apply-failure refusal, and the identity-DIFFERS branch (a
  successful `git apply` implies identity; no honest fixture reaches
  it). Stated, not checked.
- **Platform marker:** the module carries the repo's standing
  `skipif(win32 or no bash)` for bash helpers (Windows lanes' `bash`
  is the WSL stub). On macOS/Linux every case runs; nothing is marked
  skip or xfail.
- No CHANGELOG entry — `scripts/` is not shipped code.
- The pinned pre-commit run created this worktree's `.venv` via
  `uv run --frozen` (git-ignored). Nothing else on disk changed.

### DIFF
--- a/scripts/salvage_worktree_diff.sh
+++ b/scripts/salvage_worktree_diff.sh
@@ -1,0 +1,185 @@
+#!/usr/bin/env bash
+# salvage_worktree_diff.sh — read-only salvage of a stranded worktree's
+# uncommitted diff onto a fresh branch in the CURRENT checkout.
+#
+# Usage: scripts/salvage_worktree_diff.sh <source-worktree> <new-branch> \
+#            [--base <ref>]            (default base: origin/main)
+#
+# Mechanizes the recovery recipe that ran by hand four times on
+# 2026-09-11 (#2516–#2519): a session's worktree is left holding real,
+# uncommitted work. Rather than committing INTO that worktree (another
+# agent's, possibly still open), the diff is lifted verbatim onto a
+# fresh branch cut from current main, and the source worktree stays
+# untouched as the backup until the PR merges.
+#
+#   1. Refuse to run inside the source worktree, or from a dirty
+#      checkout. Everything about the source is READ-ONLY: the script
+#      never writes there (GIT_OPTIONAL_LOCKS=0 keeps git from even
+#      refreshing the source's index).
+#   2. Capture the tracked diff — `git diff --binary HEAD`, staged AND
+#      unstaged, so nothing a session `git add`ed is dropped — and the
+#      untracked file list. Nothing to salvage -> exit 2.
+#   3. Fetch the base's remote branch, then `git checkout -b <branch>
+#      <base>` here.
+#   4. `git apply --check`, then apply; copy each untracked file to the
+#      same relative path (never over a file that already exists here).
+#   5. Identity receipt: the diff here must equal the captured patch
+#      with `index` lines and `@@` hunk headers ignored. Then a receipt
+#      of files, +/- counts, untracked copies, and the next steps.
+#
+# Committed-but-unpushed work on the source branch is NOT salvaged —
+# this handles the uncommitted diff only; cherry-pick commits yourself.
+#
+# Exit status:   0  branch created, patch applied, identity held
+#                1  refused — inside the source, dirty checkout, base
+#                   fetch/checkout failed, patch does not apply, an
+#                   untracked file collides with a file here, or the
+#                   identity receipt differs
+#                2  nothing to salvage (no tracked diff, no untracked)
+#               64  usage
+set -euo pipefail
+
+usage() {
+    echo "usage: $0 <source-worktree> <new-branch> [--base <ref>]" >&2
+    exit 64
+}
+
+refuse() {
+    echo "refused: $*" >&2
+    exit 1
+}
+
+SRC=""
+BRANCH=""
+BASE="origin/main"
+while [ $# -gt 0 ]; do
+    case "$1" in
+        --base)
+            [ $# -ge 2 ] || usage
+            BASE="$2"
+            shift 2
+            ;;
+        --base=*)
+            BASE="${1#--base=}"
+            shift
+            ;;
+        -*)
+            usage
+            ;;
+        *)
+            if [ -z "$SRC" ]; then
+                SRC="$1"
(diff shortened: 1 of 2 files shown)


silversurfer562:
Adversarial verification (same-model, workflow lane) REFUTED the PR's identity claim: a source with a `git add`-ed new file or a `git mv` rename applied cleanly and then exited 1 with DIFFERS, because `git apply` ran without `--index` and the new paths stayed untracked, invisible to the destination's `git diff --binary HEAD`. Fixed in `47d17f626` (signed, fast-forward): `git apply --index` so applied changes arrive STAGED (header, receipt and the DIFFERS message say so); fixtures now add a staged new file, a `git mv` rename and a tracked BINARY edit to the source, compare `diff --binary HEAD` on both sides and assert the staged status here. Mutation-checked: dropping `--index` fails the happy path; dropping `--binary` fails it too (the verifier's should-fix — previously guarded by prose only). 10 passed serially; shellcheck clean; pinned hooks pass. Not addressed, disclosed: the DIFFERS branch has no honest fixture after this fix; `git rm --cached` + same-name untracked rewrite is refused as a collision (safe direction). Head moved from 5691d1790 → 47d17f626, so the earlier merge word needs re-giving against the new head (D10).

